Issue #1, opened 2019-08-05
Lift off


bbatsov:
Here's a rundown of the things we need to do:

- [ ] Find some maintainers for the project
- [ ] Create a style guide documenting the best MiniTest practices
- [ ] Setup the basic project structure (it's should be based on the existing extensions like `rubocop-rspec` and `rubocop-rails`
- [ ] Hack and have fun

If you want to work on all/any of this - comment on this issue. 😉

prathamesh-sonpatki:
I will take a stab at `Setup the basic project structure (it's should be based on the existing extensions like rubocop-rspec and rubocop-rails`

coding-bunny:
As mentioned on Twitter, I'd be happy to help out on this project.
Looks like a nice step up from using RuboCop and reporting issues :D

Springok:
We are using mintiest in our team, maybe I can help on the item about style guide, and I am also interested in setup of basic project structure, will check the ‘rubicon-rails’ first~

bbatsov:
Btw - I've created an empty https://github.com/rubocop-hq/minitest-style-guide

koic:
> Setup the basic project structure (it's should be based on the existing extensions like `rubocop-rspec` and `rubocop-rails`

One point. RuboCop minitest tests should be written by minitest, not RSpec. This makes RuboCop minitest itself dogfooding for RuboCop minitest.

Therefore basic project structure should be like that.

coding-bunny:
Do we have any guide for minitest to start from?
Would be cool to identify some basic cops to create or something to work towards to

bbatsov:
@coding-bunny I'm counting on those of you who use Minitest to document the best practices you know here https://github.com/rubocop-hq/minitest-style-guide I haven't used it myself in ages, so I can't help much there.

coding-bunny:
Okay, so we're pretty much on a clean slate :)
I think we should stick to "clean" minitests approaches? The reason I'm asking is that at our company we usually include/enforce the usage of `minitest-spec-rails`, but I think if we want to write cops for that stuff specifically (mixture of minitest and rspec) we'll need a separate rubocop plugin.

So how shall we best start on this? Make PRs to update the style-guide first and establish what the best-practises are for everyone?

bbatsov:
Yeah, I'd stick to the "clean" approaches and avoid framework-specific stuff.

Springok:
> Do we have any guide for minitest to start from?
[...]

PR #2, opened 2019-08-06
Remove rspec and add minitest

abhaynikam:
Related to: https://github.com/rubocop-hq/rubocop-minitest/issues/1#issuecomment-518548748

PR fixes: 
- Removes rspec and initial specs added
- Setups minitest and migrates the specs to minitest.

Screenshot: 
rubocop-minitest


koic:
We can remove .rspec file.
https://github.com/rubocop-hq/rubocop-minitest/blob/master/.rspec

abhaynikam:
@koic : Done with the changes.

Rakefile
```diff
@@ -1,6 +1,8 @@
-require "bundler/gem_tasks"
```
koic: It doesn't depend on RSpec, so it doesn't seem to need to be removed.

rubocop-minitest.gemspec
```diff
@@ -25,5 +25,5 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler", "~> 2.0"
   spec.add_development_dependency "rake", "~> 10.0"
-  spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency "minitest", "~> 5.11"
```
koic: The latest version of minitest is specified.
https://rubygems.org/gems/minitest/versions
Is there any reason not to support older versions?

abhaynikam: No reason. I just picked it from the rubygems. Can you let me know which version we should go with? I can see Rails 4.0 has minitest `4.7.5`

koic: Thanks! Let's see how it goes :-)

abhaynikam:
Thanks 🎉

PR #5, opened 2019-08-07
Add run minitest step to the circleci

abhaynikam:
Run testcases on circleci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,9 @@ jobs:
     steps:
       - checkout
       - run: bundle install
-      - run: bundle exec rake
+      - run:
+          name: Run Minitest
+          command: bundle exec rake test
 
 workflows:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -18,7 +18,9 @@ jobs:
     steps:
       - checkout
       - run: bundle install
-      - run: bundle exec rake
+      - run:
+          name: Run Minitest
+          command: bundle exec rake test
```
koic: This change is unnecessary because tests and static analysis and others are performed by default rake task.
https://github.com/rubocop-hq/rubocop-minitest/blob/e239e39e5a2f42b8748a794e4cd2294a5571445e/Rakefile#L32-L37

abhaynikam: Oh. I didn't see that. My bad.

PR #6, opened 2019-08-15
Include a Cop to enforce assert_nil over assert_equal(nil)

duduribeiro:
Hey folks 👋

`assert_equal(nil, something)` and enforces the usage of
`assert_nil(something)` instead.

I don't know if the cop that I wrote is right and maybe it needs some adjustment 😂

I also needed to copy some code from the rubocop gem itself, because the matchers itself relies on `RSpec`. Maybe we can change it in rubocop gem to share part of the code that can be reusable, and add the matchers for rspec and minitest.


lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,38 @@
...
+    #
+    # @example
+    #   # bad
+    #   assert_equal(nil, actual)
+    #
+    #   # good
+    #   assert_nil(actual)
+    #
+    class AssertNil < Cop
+      MSG = 'Prefer using `assert_nil(..)` over `assert_equal(nil, ..)`.'
```
koic: Can you replace the `..` argument with an actual value? The following address is an example.
https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/big_decimal_new.rb#L18-L19

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      def_node_matcher :assert_equal_with_nil, <<-PATTERN
+        (send nil? :assert_equal
+            nil
+            _
```
koic: `NodePattern` can get an actual value matched by `$_`.
https://github.com/rubocop-hq/rubocop/blob/v0.74.0/lib/rubocop/node_pattern.rb#L38

test/rubocop/cop/minitest/assert_nil_test.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class AssertNilTest < Minitest::Test
+  def setup
+    @cop = RuboCop::Cop::AssertNil.new
+  end
+
+  def test_adds_offense_for_use_of_assert_equal_nil
+    assert_offense(<<-RUBY, @cop)
+      class FooTest < Minitest::Test
+        def test_#bar
```
koic: `def test_dosomething` may be simpler as an example.

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,38 @@
...
+    #   # bad
+    #   assert_equal(nil, actual)
+    #
+    #   # good
+    #   assert_nil(actual)
+    #
+    class AssertNil < Cop
+      MSG = 'Prefer using `assert_nil(..)` over `assert_equal(nil, ..)`.'
+
+      def_node_matcher :assert_equal_with_nil, <<-PATTERN
```
koic: Please use squiggly heredoc in this PR.

```diff
-def_node_matcher :assert_equal_with_nil, <<-PATTERN
+def_node_matcher :assert_equal_with_nil, <<~PATTERN
```

koic:
> I also needed to copy some code from the rubocop gem itself, because the matchers itself relies on RSpec.

Awesome work! It ’s a great first step. We can hack in the future based on the code base. 👍

koic:
Can you add the setting of this cop to config/default.yml? The following is an example and let's set `VersionAdded: '0.1.0'`.
https://github.com/rubocop-hq/rubocop-rails/blob/v2.3.0/config/default.yml#L58-L61

test/test_helper.rb
```diff
@@ -1,5 +1,88 @@
...
+  end
+
+  corrector =
+    RuboCop::Cop::Corrector.new(@processed_source.buffer, cop.corrections)
+  new_source = corrector.rewrite
+
+  assert_equal(correction, new_source)
+end
+
+def assert_no_offense(source, cop, file = nil)
```
koic: I think this name of API is `assert_no_offenses` :-)

duduribeiro:
hey @koic 👋 

I did update the code with the suggested changes. 

I changed the node_matcher pattern to capture the argument: https://github.com/rubocop-hq/rubocop-minitest/pull/6/files#diff-4e8bec16638093f13a99617349303b4bR18

can you check if it is valid? thanks so much!

☕️

koic:
The `def_node_matcher` pattern looks good, but false positives occur when writing a message to `assert_equal`.

```ruby
assert_equal(nil, actual, 'message')
```

https://github.com/seattlerb/minitest/blob/ab39d35fb4e84eb866ed9c4ecb707cbf3889de42/lib/minitest/assertions.rb#L192

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
```
koic: Please add `module Minitest` as the namespace.

test/rubocop/cop/minitest/assert_nil_test.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class AssertNilTest < Minitest::Test
+  def setup
+    @cop = RuboCop::Cop::AssertNil.new
```
koic: Please add the namespace `Minitest`.

```ruby
@cop = RuboCop::Cop::Minitest::AssertNil.new
```

duduribeiro:
hey @koic 👋 

I updated the code. WDYT with the changes ?

Now the CI is failing for the missing documentation, but when I run the task it fails with:

```
rake aborted!
RuboCop::NodePattern::Invalid: Couldn't compile due to invalid token "?". Pattern: (send nil? :assert_equal nil (... $_) (str $_)? )
```

any ideas?

thanks 

☕️

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,49 @@
...
+      #
+      #   # good
+      #   assert_nil(actual)
+      #   assert_nil(actual, 'the message')
+      #
+      class AssertNil < Cop
+        MSG_WITHOUT_WARNING = 'Prefer using `assert_nil(%<argument>s)` over ' \
+        '`assert_equal(nil, %<argument>s)`.'
+        MSG_WITH_WARNING = 'Prefer using `assert_nil(%<argument>s, %<msg>s)` over ' \
+        '`assert_equal(nil, %<argument>s, %<msg>s)`.'
```
koic: I think there are other approaches, but this `MSG_*` can be combined into one as follows.

```diff
 lib/rubocop/cop/minitest/assert_nil.rb | 22 +++++++++++-----------
 1 file changed, 11 insertions(+), 11 deletions(-)

diff --git a/lib/rubocop/cop/minitest/assert_nil.rb b/lib/rubocop/cop/minitest/assert_nil.rb
index e8dcde2..a84e4eb 100644
--- a/lib/rubocop/cop/minitest/assert_nil.rb
+++ b/lib/rubocop/cop/minitest/assert_nil.rb
@@ -15,10 +15,8 @@ module RuboCop
       #   assert_nil(actual, 'the message')
       #
       class AssertNil < Cop
-        MSG_WITHOUT_WARNING = 'Prefer using `assert_nil(%<argument>s)` over ' \
-        '`assert_equal(nil, %<argument>s)`.'
-        MSG_WITH_WARNING = 'Prefer using `assert_nil(%<argument>s, %<msg>s)` over ' \
-        '`assert_equal(nil, %<argument>s, %<msg>s)`.'
+        MSG = 'Prefer using `assert_nil(%<arguments>s)` over ' \
+              '`assert_equal(nil, %<arguments>s)`.'

         def_node_matcher :assert_equal_with_nil, <<~PATTERN
           (send nil? :assert_equal nil (... $_) (str $_)? )
@@ -26,13 +24,15 @@ module RuboCop

         def on_send(node)
           assert_equal_with_nil(node) do |argument, assertion_message|
-            message = if assertion_message.empty?
-                        format(MSG_WITHOUT_WARNING, argument: argument)
-                      else
-                        format(MSG_WITH_WARNING, argument: argument,
-                                                 msg: "'#{assertion_message.first}'")
-                      end
-            add_offense(node, message: message)
+            assertion_message = if assertion_message.first.is_a?(String)
+                                  "'#{assertion_message.first}'"
+                                else
+                                  assertion_message.first
+                                end
+
+            arguments = [argument, assertion_message].compact.join(', ')
+
+            add_offense(node, message: format(MSG, arguments: arguments))
           end
         end
```

config/default.yml
```diff
@@ -0,0 +1,4 @@
+Minitest/AssertNil:
+  Description: 'Check if your test uses assert_nil instead of assert_equal(nil, something).'
```
koic: Enclose the code with backticks.

```diff
-'Check if your test uses assert_nil instead of assert_equal(nil, something).'
+'Check if your test uses `assert_nil` instead of `assert_equal(nil, something)`.'
```

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # This cop checks uses of `assert_equal(nil, ..)`
```
koic: The same description as `Description` of config/default.yml is a little more detailed.

config/default.yml
koic: Please add an URL of `StyleGuide`.

```diff
 Description: 'Check if your test uses `assert_nil` instead of `assert_equal(nil, something)`.
+StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-nil
 Enabled: true
```

lib/rubocop/cop/minitest/assert_nil.rb
duduribeiro: Thanks so much. Yeah, I did the update and it is more clear now.

duduribeiro:
hey @koic, I did apply your changes, can you review? thanks so much

bbatsov:
@duduribeiro I see all tests are failing due to passing the wrong number of arguments to one method.

duduribeiro:
@bbatsov and @koic I rebased with master and it fixed.. thanks

test/rubocop/cop/minitest/assert_nil_test.rb
```diff
@@ -0,0 +1,57 @@
...
+
+require 'test_helper'
+
+class AssertNilTest < Minitest::Test
+  def setup
+    @cop = RuboCop::Cop::Minitest::AssertNil.new
+  end
+
+  def test_adds_offense_for_use_of_assert_equal_nil
+    assert_offense(<<-RUBY, @cop)
```
koic: Can you please use squiggly heredoc?

```diff
diff --git a/test/rubocop/cop/minitest/assert_nil_test.rb b/test/rubocop/cop/minitest/assert_nil_test.rb
index 7bdc987..25fceee 100644
--- a/test/rubocop/cop/minitest/assert_nil_test.rb
+++ b/test/rubocop/cop/minitest/assert_nil_test.rb
@@ -8,7 +8,7 @@ class AssertNilTest < Minitest::Test
   end

   def test_adds_offense_for_use_of_assert_equal_nil
-    assert_offense(<<-RUBY, @cop)
+    assert_offense(<<~RUBY, @cop)
       class FooTest < Minitest::Test
         def test_do_something
           assert_equal(nil, somestuff)
@@ -17,7 +17,7 @@ class AssertNilTest < Minitest::Test
       end
     RUBY

-    assert_correction(<<-RUBY, @cop)
+    assert_correction(<<~RUBY, @cop)
       class FooTest < Minitest::Test
         def test_do_something
           assert_nil(somestuff)
@@ -27,7 +27,7 @@ class AssertNilTest < Minitest::Test
   end

   def test_adds_offense_for_use_of_assert_equal_nil_with_message
-    assert_offense(<<-RUBY, @cop)
+    assert_offense(<<~RUBY, @cop)
       class FooTest < Minitest::Test
         def test_do_something
           assert_equal(nil, somestuff, 'the message')
@@ -36,7 +36,7 @@ class AssertNilTest < Minitest::Test
       end
     RUBY

-    assert_correction(<<-RUBY, @cop)
+    assert_correction(<<~RUBY, @cop)
       class FooTest < Minitest::Test
         def test_do_something
           assert_nil(somestuff, 'the message')
@@ -46,7 +46,7 @@ class AssertNilTest < Minitest::Test
   end

   def test_does_not_offend_if_using_assert_nil
-    assert_no_offenses(<<-RUBY, @cop)
+    assert_no_offenses(<<~RUBY, @cop)
       class FooTest < Minitest::Test
         def test_do_something
           assert_nil(somestuff)
```

duduribeiro:
@koic I updated the code

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,49 @@
...
+      #   # good
+      #   assert_nil(actual)
+      #   assert_nil(actual, 'the message')
+      #
+      class AssertNil < Cop
+        MSG = 'Prefer using `assert_nil(%<arguments>s)` over ' \
+              '`assert_equal(nil, %<arguments>s)`.'
+
+        def_node_matcher :assert_equal_with_nil, <<~PATTERN
+          (send nil? :assert_equal nil (... $_) (str $_)? )
```
koic: I noticed two points of concern in the node pattern.

First, false negatives occur when a method is called as follows:

```ruby
assert_equal(nil, obj.do_something, 'message')
```

A test case will be added for the cop and the following changes will be required.

```diff
--- a/lib/rubocop/cop/minitest/assert_nil.rb
+++ b/lib/rubocop/cop/minitest/assert_nil.rb
@@ -19,7 +19,7 @@ module RuboCop
               '`assert_equal(nil, %<arguments>s)`.'

         def_node_matcher :assert_equal_with_nil, <<~PATTERN
-          (send nil? :assert_equal nil (... $_) (str $_)? )
+          (send nil? :assert_equal nil $_ (str $_)? )
         PATTERN

         def on_send(node)
@@ -30,7 +30,7 @@ module RuboCop
                                   assertion_message.first
                                 end

-            arguments = [argument, assertion_message].compact.join(', ')
+            arguments = [argument.source, assertion_message].compact.join(', ')

             add_offense(node, message: format(MSG, arguments: arguments))
           end
```

Second, the third argument of `assert_equal` can accept anything other than a string.
For example, there are false negatives in the following cases.

```ruby
assert_equal(nil, obj, message)
assert_equal(nil, obj, MESSAGE)
```

The node pattern will also need to be changed.

Can you change this?

test/rubocop/minitest_test.rb
```diff
@@ -6,8 +6,4 @@ class MinitestTest < Minitest::Test
   def test_rubocop_minitest_gem_has_a_version_number
     refute_nil RuboCop::Minitest::VERSION
   end
-
-  def test_rubocop_minitest_does_something_useful
-    assert true
-  end
```
koic: Yeah, this file can be removed.

test/rubocop/cop/minitest/assert_nil_test.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        def test_do_something
+          assert_nil(somestuff, MESSAGE)
+        end
+      end
+    RUBY
+  end
+
+  def test_does_not_offend_if_using_assert_nil
+    assert_no_offenses(<<-RUBY, @cop)
```
koic: Can you use squiggly heredoc (`<<~`)?

config/default.yml
```diff
@@ -0,0 +1,5 @@
+Minitest/AssertNil:
```
koic: It seems that only `test` directory can be targeted.

```diff
+AllCops:
+  Include:
+    - '**/test/**/*'
+
 Minitest/AssertNil:
   Description: 'Check if your test uses `assert_nil` instead of `assert_equal(nil, something)`.'
```

test/test_helper.rb
```diff
@@ -1,5 +1,88 @@
 # frozen_string_literal: true
 
-require 'rubocop/minitest'
+require 'rubocop'
+require 'rubocop-minitest'
 require 'minitest/autorun'
 require 'minitest/pride'
+
+# Lazyness copied from rubocop source code
```
koic: `Lazyness` is a misspelling of `Laziness`.

lib/rubocop/cop/minitest/assert_nil.rb
```diff
@@ -0,0 +1,45 @@
...
+      class AssertNil < Cop
+        MSG = 'Prefer using `assert_nil(%<arguments>s)` over ' \
+              '`assert_equal(nil, %<arguments>s)`.'
+
+        def_node_matcher :assert_equal_with_nil, <<~PATTERN
+          (send nil? :assert_equal nil $_ $...)
+        PATTERN
+
+        def on_send(node)
+          assert_equal_with_nil(node) do |argument, assertion_message|
```
koic: The matched pattern may be concrete if block variable names are as follows. 
e.g: 

```diff
-assert_equal_with_nil(node) do |argument, assertion_message|
+assert_equal_with_nil(node) do |actual, message|
```

duduribeiro:
@koic I changed with your suggestions. Thanks so much

config/default.yml
```diff
@@ -0,0 +1,9 @@
+AllCops:
+  Include:
+    - '**/test/**/*'
+
+Minitest/AssertNil:
+  Description: 'Check if your test uses `assert_nil` instead of `assert_equal(nil, something)`.'
+  StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-nil'
+  Enabled: true
+  VersionAdded: '0.1.0'
```
koic: I have one (probably last) feedback for this PR. The other day my comment was wrong 💦  `VersionAdded` does not include tiny version. Can you update to `VersionAdded: '0.1'`?

duduribeiro: hey @koic. I did update it.

duduribeiro:
@koic thanks for your help!

koic:
@bbatsov I'd like to release RuboCop Minitest 0.1.0 after merging this PR and maintaining some documentation. This cop is the first cop that will be published on RuboCop Minitest.
This is a great achievement by @duduribeiro.

bbatsov:
@koic I completely agree. Let's make this happen. 😄

PR #9, opened 2019-09-04
Added a cop to enforce assert_includes over assert(collection.includes?(actual))

abhaynikam:
PR adds cops to enforce the use of `assert_includes` over `assert(collection.includes?(actual))`

Ref: https://github.com/rubocop-hq/minitest-style-guide#assert-includes


koic:
Can you add the changelog entry and resolve conflicting files?

abhaynikam:
@koic Yep. I'll do it today.

PR #11, opened 2019-09-06
Add new cop `Minitest/RefuteNil`

tejasbubane:
* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## master (unreleased)
 
+### New features
+
+* [#11](https://github.com/rubocop-hq/rubocop-minitest/pull/11): Add new `Minitest/RefuteNil` cop. ([@tejasbubane ][])
+
 ## 0.1.0 (2019-09-01)
 
 ### New features
@@ -9,3 +13,4 @@
 
 [@koic]: https://github.com/koic
 [@duduribeiro]: https://github.com/duduribeiro
+[@tejasbubane]: https://github.com/tejasbubane
--- a/config/default.yml
+++ b/config/default.yml
@@ -7,3 +7,9 @@ Minitest/AssertNil:
   StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-nil'
   Enabled: true
   VersionAdded: '0.1'
+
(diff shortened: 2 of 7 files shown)


tejasbubane:
I created a separate cop for now but should we have separate refute cops or have single cop which handles refute as well? This will apply to almost all other cops.

koic:
Ah, that's a good point of view. RuboCop Minitest can be designed to provide a single cop that bundles `refute_*`.

However, a design that provides individual cops along The Minitest Style Guide may be fine. So, Minitest is a simple API and expects a limited number of style rules.

I think that it may be simple if a rule of Minitest Style Guide and a cop is mapped. Each cop can have a reference to a rule of The Minitest Style Guide.

Cc @rubocop-hq/rubocop-core

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## master (unreleased)
 
+* [#11](https://github.com/rubocop-hq/rubocop-minitest/pull/11): Add new `Minitest/RefuteNil` cop. ([@tejasbubane ][])
```
koic: Can you add `### New features` category?

koic:
Please add this cop setting to config/default.yml.

CHANGELOG.md
tejasbubane: added

koic:
Thanks!

config/default.yml
```diff
@@ -7,3 +7,9 @@ Minitest/AssertNil:
   StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-nil'
   Enabled: true
   VersionAdded: '0.1'
+
+Minitest/RefuteNil:
+  Description: 'Check if your test uses `refute_nil` instead of `refute_equal(nil, something)`.'
+  StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#refute-nil'
+  Enabled: true
+  VersionAdded: '0.1.1'
```
koic: I overlooked the version and updated it by https://github.com/rubocop-hq/rubocop-minitest/commit/b9c8da6. FYI, Minor version will be bumped when new cop are added.

PR #12, opened 2019-09-11
Modify `Minitest/AssertTruthy` to check for `refute_equal`

tejasbubane:
Alternative to separate cop approach in https://github.com/rubocop-hq/rubocop-minitest/pull/11

* [x] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,6 +19,7 @@ Naming/PredicateName:
     - def_node_search
 
 Style/FormatStringToken:
+  Enabled: false
   # Because we parse a lot of source codes from strings. Percent arrays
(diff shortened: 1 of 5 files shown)


koic:
`refute` method is a method to assert that it is not truthy.  I think `refute(true, actual)` is not practical issue.  That would mean `refute(false, actual)`.
https://github.com/seattlerb/minitest/blob/ab39d35fb4e84eb866ed9c4ecb707cbf3889de42/lib/minitest/assertions.rb#L558-L564

So, I think that `Minitest /AssertTruthy ` cop should not support `refute(true, actual)`.

bbatsov:
I agree. We should keep those cops focused on one pair of assertions each.

PR #16, opened 2019-09-27
Fixed the test description typo from assert_nil -> assert_includes

abhaynikam:
Assert Include cop test description had a typo. Changed it from `assert_nil` -> `assert_includes`


koic:
Thanks!

PR #17, opened 2019-09-27
Add cop to enfore refute_empty over refute(actual.empty?)

abhaynikam:
PR adds cops to enforce the use of refute_empty over refute(actual.empty?)

Ref: https://github.com/rubocop-hq/minitest-style-guide#refute-empty


lib/rubocop/cop/minitest/refute_empty.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Check if your test uses `refute_empty` instead of `refute(actual.empty?)`.
+      #
+      # @example
+      #   # bad
+      #   assert(actual.empty?)
```
duduribeiro: I think this doc is not equal the behave you implemented 🤔 

In the code you are registering an offsense if using `refute(actual.empty?)` and here you are saying that the bad is `assert(actual.empty?)`. 

Also, it would be awesome if you can make the cop look for both: `refute(something.empty?)` and assert(!something.empty)`.

abhaynikam: @duduribeiro: My bad. I'll fix the documentation. Regarding adding `assert(!something.empty)`, I was planning to do it in another PR. Hope that is fine

abhaynikam:
Reopened another PR: #20 for same. Sorry about this.

PR #18, opened 2019-09-27
Include a Cop to enforce refute(actual) over assert_equal(false)

duduribeiro:
`assert_equal(false, something)` and enforces the usage of
`refute(something)` instead.

-----------------

* [x] Added tests.


koic:
Can you add the changelog entry about this one?

duduribeiro:
@koic sorry for the late response. I did add the entry into the changelog. thanks so much

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 * [#19](https://github.com/rubocop-hq/rubocop-minitest/pull/19): Fix a false negative for `Minitest/AssertIncludes` when using `include` method in arguments of `assert` method. ([@abhaynikam][])
 
+* [#18](https://github.com/rubocop-hq/rubocop-minitest/pull/18): Add new `Minitest/RefuteFalse` cop. ([@duduribeiro][])
```
bbatsov: This should be under `New features`.

lib/rubocop/cop/minitest/refute_false.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Check if your test uses `refute(actual)`
```
bbatsov: The wording here is confusing. I'd write it the other way around or "This cop enforces the usages of `refute` over `assert...`. 

I'd also use `object` instead of `actual` here give there's nothing to compare against.

lib/rubocop/cop/minitest/refute_false.rb
```diff
@@ -0,0 +1,46 @@
...
+      # instead of `assert_equal(false, actual)`.
+      #
+      # @example
+      #   # bad
+      #   assert_equal(false, actual)
```
bbatsov: Shouldn't we cover `nil` as well?

test/rubocop/cop/minitest/refute_false_test.rb
```diff
@@ -0,0 +1,120 @@
...
+    assert_correction(<<~RUBY, @cop)
+      class FooTest < Minitest::Test
+        def test_do_something
+          refute(obj.do_something, 'the message')
+        end
+      end
+    RUBY
+  end
+
+  def test_registers_offense_when_using_assert_equal_with_variable_message
```
bbatsov: Testing with a variable or a constant seems redundant to me - you should only care about the number of params to infer that something's a message.

lib/rubocop/cop/minitest/refute_false.rb
herwinw: `assert_equal(nil, object)` will trigger another cop to change it to `assert_nil(object)`. Adding it here would only add confusion.

duduribeiro:
@bbatsov, @koic I did update the branch with the changes suggested here. Thanks

bbatsov:
Thanks!

PR #19, opened 2019-10-01
Fixed typo for AssertIncludes cop. Changed includes -> include

abhaynikam:
This PR fixes the typo for AssertIncludes cop. Changes `collection.includes` -> `collection.include` at all places.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#19](https://github.com/rubocop-hq/rubocop-minitest/pull/19): Fix a false negative for `Minitest/AssertIncludes` when using `include` method in arguments of `assert` method. ([@abhaynikam][])
+
 ## 0.2.1 (2019-09-24)
 
 ### Bug fixes
--- a/config/default.yml
+++ b/config/default.yml
@@ -15,7 +15,7 @@ Minitest/AssertEmpty:
   VersionAdded: '0.2'
 
 Minitest/AssertIncludes:
-  Description: 'Check if your test uses `assert_includes` instead of `assert(collection.includes?(actual))`.'
+  Description: 'Check if your test uses `assert_includes` instead of `assert(collection.include?(actual))`.'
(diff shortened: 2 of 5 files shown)


koic:
The change fixes a false negative. Can you please add the changelog entry?

abhaynikam:
@koic : Added the changelog. Can you please have a look?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#19](https://github.com/rubocop-hq/rubocop-minitest/pull/19): Fixes typo `collection.includes?` to `collection.include?` in `Minitest/AssertIncludes` cop. ([@abhaynikam][])
```
koic: It would be better to write a false negative correction rather than a typo :-) The following is an example.

```
Fix a false negative for `Minitest/AssertIncludes` when using `include` method in arguments of `assert` method.
```

abhaynikam: Thanks, @koic 🎉 . Updated the changelog

bbatsov:
Btw, I wonder if we should check for both for the sake of Rails users. I know the ActionSupport aliases are quite popular there. @koic any thoughts?

koic:
It seems my understanding is lacking 💦 Which Active Support API is that?

bbatsov:
Hmm, that's weird - I can't find it here https://edgeguides.rubyonrails.org/active_support_core_extensions.html#extensions-to-enumerable I could swear that back in the day Rails shipped some aliases like `includes?`, `contains?`, etc. Perhaps they were removed at some point?

koic:
@bbatsov Thanks for the answer. AFAIK, there is an `in?` method that works like `include?` method.
https://guides.rubyonrails.org/active_support_core_extensions.html#in-questionmark

Other methods cannot be remembered 😅 First of all, I think that it is better to support only `include?` method respond by Minitest.
https://github.com/seattlerb/minitest/blob/930ec0ba2e3ca010cca388a0429b33fd63c7d0bd/lib/minitest/assertions.rb#L239-L248

bbatsov:
Yeah, I agree.

PR #21, opened 2019-10-03
Include a Cop to enforce refute_equal over assert(negations)

duduribeiro:
`assert(some negations)` and enforces the usage of
`refute_eqaul(expected, actual)` instead.


lib/rubocop/cop/minitest/refute_equal.rb
```diff
@@ -0,0 +1,64 @@
...
+      #
+      #   # good
+      #   refute_equal("rubocop-minitest", actual)
+      #
+      class RefuteEqualNil < Cop
+        MSG = 'Prefer using `refute_equal(%<preferred>s)` over ' \
+              '`assert(%<over>s)`.'
+
+
+        def_node_matcher :assert_not_equal_operator, <<~PATTERN
```
duduribeiro: @koic I tried to isolate the both node_matchers into one but I couldn't.. do you have some idea on how can I do it?

koic: I think that can be implemented by `{` ... `}` node pattern. How about that?
https://github.com/rubocop-hq/rubocop/blob/v0.75.0/lib/rubocop/node_pattern.rb#L34

> `'{send class}'      # matches (send ...) or (class ...)`

lib/rubocop/cop/minitest/refute_equal.rb
```diff
@@ -0,0 +1,88 @@
...
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Check if your test uses `assert` with negation instead
+      # of `refute_equal(expected, actual)`.
+      #
+      # @example
+      #   # bad
+      #   assert(!"rubocop-minitest".eql?(actual))
```
koic: FYI, I opened a PR https://github.com/rubocop-hq/minitest-style-guide/pull/20.
`refute_equal` method uses `==` method, not `eql?` method.
https://github.com/seattlerb/minitest/blob/15ed8e4/lib/minitest/assertions.rb#L587-L592

`==` method and `eql?` method behave differently.

duduribeiro: interesting.. so, what do you think to keep thi Cop to only check  for `assert(expected != actual)` ? 🤔

koic: Probably `assert(expected != actual)` and `assert(!expected == actual)` will be checked. Let's first look at PR rubocop-hq/minitest-style-guide#20 trends.

bbatsov:
Most of my feedback from your other PR applies here as well.

lib/rubocop/cop/minitest/refute_equal.rb
koic: FYI, https://github.com/rubocop-hq/minitest-style-guide/pull/20 has been merged.

duduribeiro:
hey @bbatsov and @koic 👋 

I did update the code. Can you folks re-review it? Thanks!

PR #20, opened 2019-10-01
Add cop to enfore refute_empty over refute(actual.empty?)

abhaynikam:
PR adds cops to enforce the use of refute_empty over refute(actual.empty?)

Ref: https://github.com/rubocop-hq/minitest-style-guide#refute-empty

PR #23, opened 2019-10-10
Renamed the assertion argument to object from actual

abhaynikam:
As per @bbatsov comment, renamed the argument where no expectation is present to `object` from `actual`. 

Ref: https://github.com/rubocop-hq/rubocop-minitest/pull/20#discussion_r331858962

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -9,13 +9,13 @@ Minitest/AssertNil:
   VersionAdded: '0.1'
 
 Minitest/AssertEmpty:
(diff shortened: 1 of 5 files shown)


bbatsov:
This has to be rebased on top of `master`. The naming convention should also be updated in the examples in the style guide.

abhaynikam:
I'll work on it tonight. Thanks for the feedback @bbatsov

abhaynikam:
@bbatsov Raised PR to update the minitest-style-guides -> https://github.com/rubocop-hq/minitest-style-guide/pull/21

PR #26, opened 2019-10-12
Fix English typo in the class documentation

duduribeiro:
Hey folks 👋

I think we have a english typo in refute_equal and refute_false class
documentation. Instead of `enforces the usages` it should be `enforces
the use`.


bbatsov:
Thanks!

PR #15, opened 2019-09-27
Add cop to enforce refute_includes over refute(collection.include?(actual))

abhaynikam:
PR adds cops to enforce the use of refute_includes over refute(collection.includes?(actual))

Ref: https://github.com/rubocop-hq/minitest-style-guide#refute-includes

PR #38, opened 2019-11-07
[Fix #32] Add new cop assert_empty_literal

### LINKED ISSUE #32: Detection of assert when assert_equal was intended

I was wondering why some unit tests weren't failing after code changes. Turns out they were written like this:
```ruby
assert([], result.errors)
```
The test should have use `assert_equal` (or maybe `assert_empty` without the first argument), but since the empty array is trueish, this test always passes, regardless of the value of `result.errors`, which is only used as the additional (optional) message parameter.

I would like Rubocop-Minitest to find this error

## Additional context

I have absolutely no idea how this error could be detected by Rubocop. In case a third parameter is added, this could be used to detect errors:
```ruby
assert([], result.errors, 'Expected no errors')
```
But this is pretty pointless, since it would result in runtime errors anyway.

If anybody has an idea on how to detect this, I could probably implement it. If it appears to be impossible, we shouldn't attempt to implement it ;)

tejasbubane:
Closes #32

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -9,6 +9,11 @@ Minitest/AssertEmpty:
   Enabled: true
   VersionAdded: '0.2'
 
+Minitest/AssertEmptyLiteral:
+  Description: 'This cop enforces the test to use `assert_empty` instead of using `assert([], object)` or `assert({}, object)`.'
+  Enabled: true
+  VersionAdded: '0.5'
+
 Minitest/AssertEqual:
   Description: 'This cop enforces the test to use `assert_equal` instead of using `assert(expected == actual)`.'
   StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-equal-arguments-order'
--- a/lib/rubocop/cop/minitest_cops.rb
+++ b/lib/rubocop/cop/minitest_cops.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'minitest/assert_empty'
+require_relative 'minitest/assert_empty_literal'
 require_relative 'minitest/assert_equal'
 require_relative 'minitest/assert_nil'
 require_relative 'minitest/assert_includes'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#32](https://github.com/rubocop-hq/rubocop-minitest/issues/32): Add new `Minitest/AssertEmptyLiteral` cop. ([@tejasbubane][])
+
 ## 0.4.1 (2019-11-10)
 
 ### Bug fixes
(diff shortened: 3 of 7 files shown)


tejasbubane:
Not sure why build failing on `ruby HEAD`. `rake generate_cops_documentation` gives no changes.

abhaynikam:
@tejasbubane : Test is supposed to fail on Ruby Head. Refer to this comment: https://github.com/rubocop-hq/rubocop-minitest/pull/15#issuecomment-536153434

koic:
Can you add the default setting for this cop to config/default.yml?

tejasbubane:
> Can you add the default setting for this cop to config/default.yml?

Done @koic 
Also added styleguide PR https://github.com/rubocop-hq/minitest-style-guide/pull/27

> Test is supposed to fail on Ruby Head. Refer to this comment: #15 (comment)

Can we mark it as `allow failure` then?

abhaynikam:
> Can we mark it as allow failure then?
How is that done on CI? I checked CI docs but I couldn't figure out the solution. You should give it a try I think 😊

tejasbubane:
> How is that done on CI? I checked CI docs but I couldn't figure out the solution. You should give it a try I think 😊

[TravisCI has option `allow_failure`](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail). But strangely I could not find any for CircleCI - https://discuss.circleci.com/t/can-we-allow-a-job-to-fail-in-a-workflow/23343/3 🤔

lib/rubocop/cop/minitest_cops.rb
```diff
@@ -7,6 +7,7 @@
 require_relative 'minitest/assert_instance_of'
 require_relative 'minitest/assert_respond_to'
 require_relative 'minitest/assert_truthy'
+require_relative 'minitest/assert_empty_literal'
```
koic: Please move this `require_relative 'minitest/assert_empty_literal'` after `require_relative 'minitest/assert_empty'` to arrange it in ASCII order.

config/default.yml
```diff
@@ -45,6 +45,12 @@ Minitest/AssertTruthy:
   Enabled: true
   VersionAdded: '0.2'
 
+Minitest/AssertEmptyLiteral:
+  Description: 'This cop enforces the test to use `assert_empty` instead of using `assert([], object)` or `assert({}, object)`.'
+  StyleGuide: 'https://github.com/rubocop-hq/minitest-style-guide#assert-empty-literal'
```
koic: I think this is a cop that acts as a Linter to prevent bugs. So I probably don't need a style guide.

lib/rubocop/cop/minitest_cops.rb
tejasbubane: changed

config/default.yml
tejasbubane: removed

koic:
CI has failed. Can you rebase with the latest master?

tejasbubane:
> CI has failed. Can you rebase with the latest master?

@koic Rebased.

koic:
Thanks!

PR #40, opened 2019-11-08
[Fix #39] Fix an incorrect autocorrect for `Minitest/AssertRespondTo`

### LINKED ISSUE #39: Minitest/AssertRespondTo autocorrect causes ArgumentError: wrong number of arguments (given 1, expected 2..3)
Given the test statement:

`assert respond_to? :token_payment_method`

The Minitest/AssertRespond cop auto corrects that to 

`assert_respond_to(:token_payment_method)`

When the test is run, an error message is returned: 

`ArgumentError: wrong number of arguments (given 1, expected 2..3)`

The log message is

`[Corrected] Minitest/AssertRespondTo: Prefer using assert_respond_to(:token_payment_method) over assert(respond_to? :token_payment_method).`

Minitest assert_respond_to requires an object:

> assert_respond_to(obj, meth, msg = nil)

This can be corrected by editing the corrected test statement to:

`assert_respond_to(self, :token_payment_method)`

Using Rails 6.01, Ruby 2.65, Minitest 5.13.0

koic:
Fixes #39.

This PR fixes an incorrect autocorrect for `Minitest/AssertRespondTo` and `Minitest/RefuteRespondTo` when using assertion method calling `respond_to` with receiver omitted.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/minitest/assert_respond_to.rb
+++ b/lib/rubocop/cop/minitest/assert_respond_to.rb
@@ -10,10 +10,12 @@ module RuboCop
       #   # bad
       #   assert(object.respond_to?(:some_method))
       #   assert(object.respond_to?(:some_method), 'the message')
+      #   assert(respond_to?(:some_method))
       #
       #   # good
       #   assert_respond_to(object, :some_method)
       #   assert_respond_to(object, :some_method, 'the message')
+      #   assert_respond_to(self, some_method)
       #
       class AssertRespondTo < Cop
         MSG = 'Prefer using `assert_respond_to(%<preferred>s)` over ' \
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/minitest/assert_respond_to.rb
```diff
@@ -38,13 +37,20 @@ def autocorrect(node)
...
+              preferred = build_preferred_arguments(object, method, custom_message)
               replacement = "assert_respond_to(#{preferred})"
               corrector.replace(node.loc.expression, replacement)
             end
           end
         end
+
+        private
+
+        def build_preferred_arguments(receiver, method, message)
```
abhaynikam: @koic Since we are allowing `assert respond_to? :some_method`. Shouldn't we add the change in the example here: https://github.com/rubocop-hq/rubocop-minitest/pull/40/files#diff-8a61a4905f40fa4acda025d5ed4bbb66R9

WDYT?

koic: It seems that `assert(respond_to?(:some_method))` does not exist in the examples and documentation. What does that mean?

abhaynikam: I wanted to ask if we should update the example and documentation as follows

```
# Good
assert_respond_to(self, some_method)

# Bad
assert(respond_to?(:some_method))
```

koic: I get it! I updated the examples. Thanks for your explanation!

PR #85, opened 2020-07-01
Add autocorrect to `Rails/AssertEmptyLiteral` cop

fatkodima:
1. Added autocorrection
2. Also this cop was not implemented correctly as it should check against `assert_equal([], something)`, not `assert([], something)` - changed that

For reference: http://docs.seattlerb.org/minitest/Minitest/Assertions.html#method-i-assert


fatkodima:
Rebased.

koic:
> 1. Added autocorrection
> 2. Also this cop was not implemented correctly as it should check against `assert_equal([], something)`, not `assert([], something)` - changed that

I think that unsafe auto-correction could be added. But the `assert` method has been intentionally used.
https://github.com/rubocop-hq/rubocop-minitest/issues/32

I'm worried about changing behaviour for bad case.

fatkodima:
> But the assert method has been intentionally used. #32

Aaaah, got it. 

Reverted the second part:
> Also this cop was not implemented correctly as it should check against assert_equal([], something), not assert([], something) - changed that

Marked unsafe autocorrection.

koic:
Thank you!

PR #88, opened 2020-07-02
Add new `Minitest/MultipleAssertions` cop

fatkodima:
This cop checks for too many assertions per test case (with configurable `Max` option).
I set initial value to 3 (not 1), as it seems more practical number, from my experience. Wdyt?

I extracted some helpers into new mixin (not happy with its name, I think there is a better name) as I will reuse them in other future cops.


abhaynikam:
I know this is a configurable cop but I was just curious to understand the reasoning behind why limit the assertions in a spec?  I mean if you look at Rails repository any specs(https://github.com/rails/rails/blob/master/actiontext/test/unit/model_test.rb#L16-L23),  you would find tons of assertion in one spec. 

Thanks 😊

config/default.yml
```diff
@@ -77,6 +77,12 @@ Minitest/GlobalExpectations:
   Enabled: true
   VersionAdded: '0.7'
 
+Minitest/MultipleAssertions:
+  Description: 'This cop checks if test cases contain too many assertion calls.'
+  Enabled: 'pending'
+  VersionAdded: '0.10'
+  Max: 3
```
abhaynikam, line 84: Is `3` the default max value when this cop is enabled? We should mention it in the documentation somewhere. Maybe here: https://github.com/rubocop-hq/rubocop-minitest/pull/88/files#diff-579fb7ac8b2c940a3dfbb36ea97f6ea1R394

bbatsov: This will appear in the auto-generated docs (the default value), but extending the cop description is never a bad idea.

bbatsov:
> I know this is a configurable cop but I was just curious to understand the reasoning behind why limit the assertions in a spec?

It's a common good practice in xUnit - keeps you from asserting all sorts of unrelated things in a single example. I know many people who swear by the policy to have a single assertion in each test case. Obviously, there are no hard and fast rules, but that's certainly something fairly common (especially in the Java world).

fatkodima:
Many people tend to think (and there are numerous posts and books about that) that test cases should have only one assertion and having > 1 - is a bad thing. But it's not, just make sure they are related and testing one thing. As one stated, `A single assert per unit test is a great way to test the reader's ability to scroll up and down`. Thats why I chose `3` as a practical default value.

But having `too many` assertions is also a problem as it complicates understanding of the test case and makes it a little difficult  to debug when something breaks. And how many is `too many` everyone decides to himself.

abhaynikam:
Awesome. Thanks for the explanation @fatkodima and @bbatsov 😊 🎉 

I still prefer to go with multiple assertions, sometimes the redundant setup happening for the small assertion feels awkward to me but I agree small assertion helps in proper debugging when something goes wrong.

koic:
For example, RSpec (not Minitest 💦 ) has a feature that bundles tests called `aggregate_failures` to a way to solve slow tests. There is a trade-off between independent testing and slow testing. However, independent testing is definitely good principle until the problem of slow testing occurs. First, let's merge and see what happens.

koic:
Thank you!

PR #89, opened 2020-07-02
Add new `Minitest/TestCaseName` cop

fatkodima:
This cop enforces that public test case method names start with `test_` prefix.

Missed `test_` prefix is almost always an unintentional mistake - as you can see from 2 corrected tests in this PR - https://github.com/rubocop-hq/rubocop-minitest/pull/89/files#diff-50588427db1de725ce02fc0d20cee56aR88 and https://github.com/rubocop-hq/rubocop-minitest/pull/89/files#diff-f64b784295c232ad3b46ae3d9a8808acR88.


lib/rubocop/cop/minitest/test_case_name.rb
```diff
@@ -0,0 +1,66 @@
...
+      #   end
+      #
+      #   # good
+      #   class FooTest < Minitest::Test
+      #     def test_does_something
+      #       assert_equal 42, do_something
+      #     end
+      #   end
+      #
+      class TestCaseName < Cop
```
koic, line 23: I think that `TestMethodName` may be clearer.

lib/rubocop/cop/minitest/test_case_name.rb
```diff
@@ -0,0 +1,66 @@
...
+        include MinitestExplorationHelpers
+        include DefNode
+
+        MSG = 'Test case name should start with `test_` prefix.'
```
koic, line 27:
```suggestion
        MSG = 'Test method name should start with `test_` prefix.'
```

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -47,6 +47,15 @@ module MinitestExplorationHelpers
         refute_same
       ].to_set.freeze
 
+      HOOKS = %i[
```
koic, line 50:
```suggestion
      LIFECYCLE_HOOK_METHODS = %i[
```

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -74,6 +83,10 @@ def test_cases(class_node)
       def assertion?(node)
         node.send_type? && ASSERTIONS.include?(node.method_name)
       end
+
+      def hook?(node)
```
koic, line 87:
```suggestion
      def lifecycle_hook_method?(node)
```

fatkodima:
Updated.

koic:
Thanks!

PR #90, opened 2020-07-02
Add new `Minitest/OverridenHook` cop

fatkodima:
This cop enforces that there is only one definition for each hook type.

This is always a mistake as it overrides previous hook method definition, which can not be detected without warnings enabled and even with them it is hard to do so. I have made myself and have seen this mistake a lot, and I guess it originates from `rspec` where it is possible to define multiple `before` and `after` hooks.


koic:
I think it is unnecessary to add this cop because it can be detected by `Lint/DuplicateMethods` of RuboCop core.

fatkodima:
Ah, forgot about its existence. I should be more careful.

PR #92, opened 2020-07-02
Add new `Minitest/LiteralAsActualArgument ` cop

fatkodima:
https://minitest.rubystyle.guide/#assert-equal-arguments-order

While, unfortunately, we can not for sure tell in every occasion if expected and actual are incorrectly swapped. But we can for sure tell so when actual is a [recursive]literal.


lib/rubocop/cop/minitest/assert_equal_arguments_order.rb
```diff
@@ -0,0 +1,41 @@
...
+      #   assert_equal foo, 2
+      #   assert_equal foo, [1, 2]
+      #   assert_equal foo, [1, 2], 'message'
+      #
+      #   # good
+      #   assert_equal 2, foo
+      #   assert_equal [1, 2], foo
+      #   assert_equal [1, 2], foo, 'message'
+      #
+      class AssertEqualArgumentsOrder < Cop
```
koic, line 20: For example, what about the name `LiteralAsActualArguemnt`?

lib/rubocop/cop/minitest/assert_equal_arguments_order.rb
```diff
@@ -0,0 +1,41 @@
...
+        MSG = 'You misordered expected and actual arguments for `assert_equal`.'
```
koic, line 21: For example, the following message may be better.

```suggestion
        MSG = 'Replace the literal with the first argument.'
```

test/rubocop/cop/minitest/assert_equal_arguments_order_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+require 'test_helper'
+
+class AssertEqualArgumentsOrderTest < Minitest::Test
+  def test_registers_offense_when_actual_is_basic_literal
+    assert_offense(<<~RUBY)
+      class FooTest < Minitest::Test
+        def test_do_something
+          assert_equal(foo, 2)
+          ^^^^^^^^^^^^^^^^^^^^ You misordered expected and actual arguments for `assert_equal`.
```
koic, line 11: I think the method name is fine. E.g.:

```suggestion
          assert_equal(foo, 2)
                       ^^^^^^ You misordered expected and actual arguments for `assert_equal`.
```

fatkodima:
Updated.

koic:
Thanks!

PR #93, opened 2020-07-03
Add new `Minitest/EmptyTestCase` cop

fatkodima:
This cop checks for empty test cases.


koic:
I think it is unnecessary to add this cop because it can be detected by `Style/EmptyMethod` cop of RuboCop core.
So same as https://github.com/rubocop-hq/rubocop-minitest/pull/94#issuecomment-653618848.

PR #94, opened 2020-07-03
Add new `Minitest/EmptyHook` cop

fatkodima:
This cop checks for empty hooks.

Follow up of #93


koic:
I think it is unnecessary to add this cop because it can be detected by `Style/EmptyMethod` cop of RuboCop core.

koic:
RuboCop Minitest should have cop specializing in Minitest. So any issues that RuboCop core can detect should be left to RuboCop core :-)

PR #96, opened 2020-07-03
Add new `Minitest/UnspecifiedException` cop

fatkodima:
To avoid false-positives when checking exceptions, it is valuable to be able to enforce that each `assert_raises` includes specific exception(s) it checks against.


docs/modules/ROOT/pages/cops_minitest.adoc
```diff
@@ -785,3 +785,30 @@ refute_respond_to(self, :do_something)
...
+
+This cop checks for a specified error in `assert_raises`.
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+assert_raises { do_something}
+assert_raises('message') { do_something}
```
koic, line 809: I think this case can be allowed like RuboCop RSpec.
https://github.com/rubocop-hq/rubocop-rspec/blob/v1.41.0/lib/rubocop/cop/rspec/unspecified_exception.rb#L24-L26

koic:
This proposal looks good to me. Can you open a PR to the style guide for this rule?

docs/modules/ROOT/pages/cops_minitest.adoc
fatkodima: Minitest does not accept strings as exception messages, only exception classes - http://docs.seattlerb.org/minitest/Minitest/Assertions.html#method-i-assert_raises

It only accepts string message as last argument.

fatkodima:
Opened https://github.com/rubocop-hq/minitest-style-guide/pull/33 and added `StyleGuide` section to that.
Changed cop class example with `'message'`, to reflect that it is not related to `Exception#message`, as opposite to rspec.

docs/modules/ROOT/pages/cops_minitest.adoc
```diff
@@ -785,3 +785,34 @@ refute_respond_to(self, :do_something)
...
+|===
+
+This cop checks for a specified error in `assert_raises`.
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+assert_raises { do_something}
```
abhaynikam, line 808: Space after `do_something` was missing. Maybe we could rename `do_something` to `do_something_that_raises_exception`. Just more readable 😊 

```suggestion
assert_raises { do_something }
```

fatkodima: Thanks! Changed it to `raise FooException`.

koic:
Thanks!

PR #99, opened 2020-07-05
Update `Minitest::Test` child check

koic:
This PR removes `MinitestExplorationHelpers#minitest_test_subclass` because For example, Rails testing does not inherit `Minitest::Test` directly but inherits `ActiveSupport::TestCase`.

A class that inherits some superclass and whose class name ends with `Test` is identified as a test class.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -49,13 +49,9 @@ module RuboCop
 
       private
 
-      def minitest_test_subclass?(class_node)
-        minitest_test?(class_node.parent_class)
+      def test_class?(class_node)
+        class_node.parent_class && class_node.identifier.source.end_with?('Test')
       end
-
-      def_node_matcher :minitest_test?, <<~PATTERN
-        (const (const nil? :Minitest) :Test)
-      PATTERN
 
       def test_cases(class_node)
         class_def = class_node.body
--- a/test/rubocop/cop/minitest/multiple_assertions_test.rb
+++ b/test/rubocop/cop/minitest/multiple_assertions_test.rb
@@ -19,7 +19,30 @@ class MultipleAssertionsTest < Minitest::Test
     RUBY
   end
 
-  def test_checks_only_minitest_test_children
+  def test_checks_when_inheriting_some_class_and_class_name_ending_with_test
+    assert_offense(<<~RUBY)
+      class FooTest < ActiveSupport::TestCase
+        def test_asserts_twice
+            ^^^^^^^^^^^^^^^^^^ Test case has too many assertions [2/1].
+          assert_equal(foo, bar)
+          assert_empty(array)
+        end
+      end
+    RUBY
+  end
+
+  def test_checks_when_inheriting_some_class_and_class_name_does_end_with_test
+    assert_no_offenses(<<~RUBY)
+      class Foo < Base
+        def test_asserts_twice
+          assert_equal(foo, bar)
+          assert_empty(array)
+        end
+      end
+    RUBY
+  end
+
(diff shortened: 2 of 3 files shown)


fatkodima:
I was also thinking about usefulness of this check and wondered if some more looser check can be made.

With proposed change, are we safe to assume, that every class inside `Include` section
```
Minitest:
  Enabled: true
  Include:
    - '**/test/**/*'
    - '**/*_test.rb'
```

is a test cases class?

Example: https://github.com/rails/rails/blob/31148cd6bef4f3a3059c51c587a8bff78a2e73e3/activerecord/test/cases/base_test.rb#L32-L56

How about checking if a class ends with `Test` (and maybe inherits something, maybe even if that something also includes `Test` in it) and assume this is a test cases class?

koic:
> How about checking if a class ends with `Test` (and maybe inherits something, maybe even if that something also includes `Test` in it) and assume this is a test cases class?

This suggestion sounds good to me. I'll update this PR later :-)

koic:
I updated this PR based on @fatkodima's suggestion.

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -50,13 +50,9 @@ module MinitestExplorationHelpers
       private
 
       def minitest_test_subclass?(class_node)
```
fatkodima, line 52:
```suggestion
      def test_class?(class_node)
```

test/rubocop/cop/minitest/multiple_assertions_test.rb
```diff
@@ -19,7 +19,30 @@ def test_asserts_twice
     RUBY
   end
 
-  def test_checks_only_minitest_test_children
+  def test_checks_when_inheriting_some_class_and_class_name_ending_with_end
```
fatkodima, line 22:
```suggestion
  def test_checks_when_inheriting_some_class_and_class_name_ending_with_test
```

test/rubocop/cop/minitest/multiple_assertions_test.rb
```diff
@@ -19,7 +19,30 @@ def test_asserts_twice
...
+        def test_asserts_twice
+            ^^^^^^^^^^^^^^^^^^ Test case has too many assertions [2/1].
+          assert_equal(foo, bar)
+          assert_empty(array)
+        end
+      end
+    RUBY
+  end
+
+  def test_checks_when_inheriting_some_class_and_class_name_does_not_test_with_end
```
fatkodima, line 34:
```suggestion
  def test_checks_when_inheriting_some_class_and_class_name_does_end_with_test
```

test/rubocop/cop/minitest/multiple_assertions_test.rb
koic: Oops! I fixed it.

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
koic: Sure!

tejasbubane:
> A class that inherits some superclass and whose class name ends with Test is identified as a test class.

Rather than assuming things, I think we should create a list of `allowed superclasses`, include Rails' default ones & allow users to configure rest.

fatkodima:
I would better stick with the cited solution, at least for first time to see how it goes. I think, our assumption is pretty solid. 

Rails has numerous base test classes and with every `Active<Whenever>` their count is only increases and this gem would need to be updated and released every time to include them. And many projects create their custom base classes and after including `rubocop-minitest` they should know/remember to include those classes before everything will start to work. Or they will see that everything is good when none of the files are checked in reality 😄

koic:
Yeah, a list method is one solution, but this time I think it can be solved by convention without using configuration. First, let's proceed with this method.

PR #105, opened 2020-07-11
Update `assertion?` check to look at method name prefix

### LINKED ISSUE #100: Relax checking of test case
Similar to https://github.com/rubocop-hq/rubocop-minitest/pull/99, I'm suggesting to identify test case as a method call, whose name starts with `assert_` or `refute_`. This will automatically work with custom assertions and rails' specific assertions.

https://github.com/rubocop-hq/rubocop-minitest/blob/85abdbea1853df82928daa90c29bb6203f418919/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb#L70-L72

If sounds good, will submit a pr.

fatkodima:
Closes #100


koic:
Thanks!

PR #112, opened 2020-10-29
Fix failing CI due to new Style/DocumentDynamicEvalDefinition cop

andyw8:
Master is currently failing due to the new `Style/DocumentDynamicEvalDefinition` cop so I've regenerated `.rubocop_todo.yml`.


koic:
Thanks!

PR #116, opened 2021-01-06
[Fix #115] Fix a false positive for `Minitest/TestMethodName`

### LINKED ISSUE #115: Minitest/TestMethodName cop auto-fixer breaks things.
The Minitest/TestMethodName cop autofixer performs a simple rename of any methods which are not lifecycle methods and which don't begin with `test_`.  This is not a safe thing to do.

Minitest already only runs methods which begin with `test_`. Everything else is likely a helper method (or a mistake). Renaming a helper method breaks everything which uses it, and creates an empty test.

--------

## Expected behavior

Only make safe changes. You could do one of the following:

* Only flag & rename methods which contain assertions.
* Check whether methods are called anywhere else when renaming them. Given that they could be called from nearly anywhere, this might be a challenge.
* Disable the auto-fixer.
* Consider whether this cop needs to exist? As I said, minitest defines tests as methods named `test_`. Is this cop meant to stop people from accidentally writing tests which don't get run, or are inline helper methods, which are specific to that test class, bad practice?

## Actual behavior

The auto-fixer renamed my helper methods, breaking my tests.

* Define a helper method within a test class. Here's an example:

```ruby

require 'test_helper'
class SomeTest < Minitest::Test
  include TestHelper

  # Define a helper method
  def tested(foo = 'useful default')
    # do stuff with foo, return test result
  end

  def test_something
    # test setup

    # use helper method:
    assert_equal 'correct answer', tested
  end
end

```

* Run `rubocop --autocorrect`, get this:

```ruby

require 'test_helper'
class SomeTest < Minitest::Test
  include TestHelper

  # Name changed here:
  def test_tested(foo = 'useful default')
    # do stuff
  end

  def test_something
    # setup

    # But not here:
    assert_equal 'correct answer', tested
  end
end

```

If you'd like a real-world example, you can use [jekyll_picture_tag](https://github.com/rbuchberger/jekyll_picture_tag). Setup instructions are pretty simple, documented here:

* [setup](https://rbuchberger.github.io/jekyll_picture_tag/devs/contributing/setup.html)
* [Tests & checks](https://rbuchberger.github.io/jekyll_picture_tag/devs/contributing/testing.html)

Basically run the tests, then autocorrect, then run the tests again. Note that currently that cop is disabled under `test/.rubocop.yml`, you'll have to re-enable it to reproduce the issue.

## RuboCop version

```
$ bundle exec rubocop -V
1.7.0 (using Parser 2.7.2.0, rubocop-ast 1.4.0, running on ruby 2.5.8 x86_64-linux)
  - rubocop-minitest 0.10.2
  - rubocop-rake 0.5.1
```

Also, I just want to say thanks. Your project has made maintaining my project much easier, and I really appreciate that!

koic:
Fixes #115.

This PR fixes a false positive for `Minitest/TestMethodName` for when defining test method takes no arguments, and test method without assertion methods.

-----------------

* [x] Added tests.

### DIFF
--- a/test/rubocop/cop/minitest/test_method_name_test.rb
+++ b/test/rubocop/cop/minitest/test_method_name_test.rb
@@ -7,10 +7,12 @@ class TestMethodNameTest < Minitest::Test
     assert_offense(<<~RUBY)
       class FooTest < Minitest::Test
         def test_do_something
+          assert_equal(expected, actual)
         end
 
         def do_something_else
             ^^^^^^^^^^^^^^^^^ Test method name should start with `test_` prefix.
+          assert_equal(expected, actual)
         end
       end
     RUBY
@@ -18,9 +20,11 @@ class TestMethodNameTest < Minitest::Test
     assert_correction(<<~RUBY)
       class FooTest < Minitest::Test
         def test_do_something
+          assert_equal(expected, actual)
         end
 
         def test_do_something_else
+          assert_equal(expected, actual)
         end
       end
     RUBY
@@ -39,6 +43,7 @@ class TestMethodNameTest < Minitest::Test
     assert_no_offenses(<<~RUBY)
       class FooTest < Minitest::Test
         def setup
+          assert_equal(expected, actual)
         end
       end
     RUBY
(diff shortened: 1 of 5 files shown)


rbuchberger:
This looks great! It definitely fixes my issues. Thanks for the quick work. 

One suggestion: You might add expectations from minitest/mock and mocha. (`.expects()`, `.expect()`, etc).

test/rubocop/cop/minitest/test_method_name_test.rb
```diff
@@ -50,6 +55,27 @@ class FooTest < Minitest::Test
         private
 
         def do_something
+          assert_equal(expected, actual)
+        end
+      end
+    RUBY
+  end
+
+  def test_does_not_register_offense_when_defining_test_method_takes_no_arguments
```
andyw8, line 64: This test name doesn't seem to match the implementation (it has an argument).

koic:
Good point! OTOH, I'm worried about the impact of not checking test method arguments (That’s currently released implementation). So first merge in this state and we can extend them (`expect` and others) later.

rbuchberger:
That makes sense! After making that comment it occurred to me that a false negative is far less harmful than a false positive.

Issue #117, opened 2021-02-03
Suggest `assert_equal` when using `assert` with multiple arguments


cstyles:
## Is your feature request related to a problem? Please describe.

We recently caught an issue where we accidentally used the following code...

```ruby
assert 1, search_table_rows.size
```

...when we really wanted this:

```ruby
assert_equal 1, search_table_rows.size
```

The first example will always pass, effectively making the test useless. This has happened a number of times before (someone even previously reported it in #32) and the only way I know of to prevent it is to hope that someone catches it in code review.

## Describe the solution you'd like

The most obvious solution is to register a violation when two arguments are passed to `assert` and to suggest `assert_equal` in that case. However, `assert` _does_ accept a second argument – a custom message to be displayed if the test fails:

https://github.com/seattlerb/minitest/blob/b8fb1014ad1932eddefb5107c0f6504ba45e0cbc/lib/minitest/assertions.rb#L178

It would be annoying to get a violation every time the user actually wanted to use `msg`. A simple improvement would be to not register a violation if the second argument is a literal `String`.

I poked around using [grep.app](https://grep.app/) and found that most usages of the `msg` argument were using a literal `String`:

https://github.com/thoughtbot/paperclip/blob/e3f7493b4cc954405a0209d26933830333ae2869/spec/paperclip/validators/attachment_file_name_validator_spec.rb#L37
https://github.com/theforeman/foreman/blob/46fcf9361e2d8c6cb89af2c3f8b9be60365fdd21/test/models/orchestration/dhcp_test.rb#L12
https://github.com/faker-ruby/faker/blob/664cbafe9d15bddb97843dcbe107204e6e7291cb/test/test_locale.rb#L75
https://github.com/Linuxbrew/legacy-linuxbrew/blob/532040391bb984e410b7ad73bdd0c518f945cc0b/Library/Formula/lcm.rb#L62
https://github.com/rails/rails/blob/da418dc2508eb8af94ba88b70034021cd17b1abd/activerecord/test/cases/validations/uniqueness_validation_test.rb#L80

Most of the instances I could find where `msg` _wasn't_ a literal `String` were in custom assertion code like the following:

https://github.com/thoughtbot/shoulda-context/blob/970d3d57a584ecb2652f0bc7188761024de16c52/lib/shoulda/context/assertions.rb#L94
[...]

PR #118, opened 2021-02-21
Change `AssertEmptyLiteral` cop to check for `assert_equal`

cstyles:
While this cop is correct that we should be using `assert_empty(list)`
over `assert([], list)`, it's right for the wrong reasons. That example
is bad because it's using `assert` and not `assert_equal`. An empty
list/hash is truthy so the assertion will always succeed. That is,
`assert([], [1, 2, 3])` will succeed even though at a glance it looks
like it shouldn't. I believe that what this cop is intending to check
for is `assert_equal([], list)` instead so I've updated the references
to `assert` to `assert_equal`.

Issue #117 proposes that we try to detect when `assert` is used where
the user likely wanted `assert_equal` so while this cop will no longer
register a violation for `assert([], list)`, if that proposal is
accepted then a future cop *will* register a violation.

edit: #120 adds such a cop.

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#118](https://github.com/rubocop-hq/rubocop-minitest/pull/118): **(BREAKING)** Fix `Minitest/AssertEmptyLiteral` by making it check for `assert_equal([], array)` instead of `assert([], array)`. ([@cstyles][])
+
 ## 0.10.3 (2021-01-12)
 
 ### Bug fixes
@@ -177,3 +181,4 @@
 [@andrykonchin]: https://github.com/andrykonchin
 [@fatkodima]: https://github.com/fatkodima
 [@tsmmark]: https://github.com/tsmmark
+[@cstyles]: https://github.com/cstyles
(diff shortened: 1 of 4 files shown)


andyw8:
cc @tejasbubane who authored this in #38

koic:
This change will be complemented by #120 and will be acceptable.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#118](https://github.com/rubocop-hq/rubocop-minitest/pull/118): Fix `Minitest/AssertEmptyLiteral` by making it check for `assert_equal([], array)` instead of `assert([], array)`. ([@cstyles][])
```
koic, line 7: This is a breaking change. Can you mark `**(Breaking)**` and squash your commits into one?

```suggestion
* [#118](https://github.com/rubocop-hq/rubocop-minitest/pull/118): **(Breaking)** Fix `Minitest/AssertEmptyLiteral` by making it check for `assert_equal([], array)` instead of `assert([], array)`. ([@cstyles][])
```

tejasbubane (approved):

CHANGELOG.md
cstyles: Fixed in 70f02e09b89d980c67a0c60af80778411fe12b90

koic:
Nicely done!

PR #121, opened 2021-02-21
Require `yaml` and `pathname` in `bin/console`

cstyles:
The absence of the `yaml` require was preventing `bin/console` from
being able to execute. Requiring `pathname` doesn't seem to be necessary
but I threw it in too for good measure.

-----------------

* [ ] Added tests.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -2,6 +2,8 @@
 # frozen_string_literal: true
 
 require 'bundler/setup'
+require 'pathname'
+require 'yaml'
 require 'rubocop/minitest'
 
 # You can add fixtures and/or initialization code here to make experimenting


andyw8 (approved):

koic:
It may be better to add `require 'rubocop'` to development script `bin/console` without affecting the runtime code.

```diff
diff --git a/bin/console b/bin/console
index 1cb4c8a..0a62a08 100755
--- a/bin/console
+++ b/bin/console
@@ -2,6 +2,7 @@
 # frozen_string_literal: true

 require 'bundler/setup'
+require 'rubocop'
 require 'rubocop/minitest'

 # You can add fixtures and/or initialization code here to make experimenting
```

cstyles:
Requiring `rubocop` increases the startup time by a lot. Here's how long it takes on this branch:

```sh
$ hyperfine --warmup 3 'bin/console -v'
Benchmark #1: bin/console -v
  Time (mean ± σ):     349.8 ms ±  12.0 ms    [User: 241.8 ms, System: 90.8 ms]
  Range (min … max):   339.4 ms … 374.6 ms    10 runs
```

Versus with `require 'rubocop'`:

```sh
$ hyperfine --warmup 3 'bin/console -v'
Benchmark #1: bin/console -v
  Time (mean ± σ):     971.4 ms ±   3.9 ms    [User: 735.0 ms, System: 217.0 ms]
  Range (min … max):   967.3 ms … 980.5 ms    10 runs
```

If that's tolerable then I'll amend the commit to use `require 'rubocop'`. Although in my opinion it's still good practice in general to list the gems you rely on in a given file, even if you know that they'll be required elsewhere first. So we could have both.

koic:
Ah, sure! How about adding these to bin/consle instead of `require 'rubocop'`?

```ruby
require 'pathname'
require 'yaml'
```

bin/console
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+require 'pathname'
+require 'yaml'
+
 require 'bundler/setup'
 require 'rubocop/minitest'
```
koic, line 8: Thank you for the update. Can you tweak it bellow?

```suggestion
require 'bundler/setup'
require 'pathname'
require 'yaml'
require 'rubocop/minitest'
```

koic:
Thanks!

PR #122, opened 2021-02-25
Fix `Minitest/TestMethodName` for tests with multiple assertions

ghiculescu:
Currently it only works for tests with a single assertion.


koic:
Thanks!

PR #124, opened 2021-03-15
Add `AnyAssertions` cop

ghiculescu:
Fixes https://github.com/rubocop/rubocop-minitest/issues/123

-----------------

* [ ] Wrote [good commit messages][1].
* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Squashed related commits together.
* [ ] Added tests.
* [ ] The PR relates to *only* one subject with a clear title
* [ ] Run `bundle exec rake default`. It executes all tests and RuboCop for itself, and generates the documentation.


config/default.yml
```diff
@@ -4,6 +4,11 @@ Minitest:
     - '**/test/**/*'
     - '**/*_test.rb'
 
+Minitest/AnyAssertions:
+  Description: 'This cop checks for at least one assertion (or flunk) in tests.'
```
andyw8, line 8:
```suggestion
  Description: 'This cop checks for at least one assertion (or refutation) in tests.'
```
I think it's better to avoid the informal 'flunk' and stay consistent with minitest's language.

andyw8:
I posted a question in the issue that I think still needs to be addressed:

> Another thing to keep in mind is that the assertion may happen in a method called by the test, in which case it wouldn't be possible to catch through static analysis.

ghiculescu:
> I posted a question in the issue that I think still needs to be addressed:
> 
> > Another thing to keep in mind is that the assertion may happen in a method called by the test, in which case it wouldn't be possible to catch through static analysis.

I don't have any good ideas for this, sorry for not addressing it directly. The `MultipleAssertions` cop would have the same problem.

I guess this is where `rubocop:disable` comes in. It doesn't feel like great practice to me to have no assertions inside your test.

config/default.yml
ghiculescu: Flunk is a minitest method: https://www.rubydoc.info/github/seattlerb/minitest/Minitest%2FAssertions:flunk

I think it's subtly different to a refutation. A flunk, to me, means "this test hasn't been written yet". Maybe other people use it differently.

andyw8: Ah, I hadn't come across that before.

config/default.yml
```diff
@@ -4,6 +4,11 @@ Minitest:
...
+  Enabled: 'pending'
```
koic, line 9: > I guess this is where `rubocop:disable` comes in. It doesn't feel like great practice to me to have no assertions inside your test.

I think this cop is worth adding to RuboCop Minitest 👍  On the other hand, assuming `rubocop:disable` pragma is not better for user experience. So, I have the same concerns as @andyw8's opinion. Can you set the default to `false`?

```suggestion
  Enabled: false
```

config/default.yml
```diff
@@ -4,6 +4,11 @@ Minitest:
...
+  VersionAdded: '0.11'
```
koic, line 10: My review was delayed and RuboCop Minitest 0.11 has been released. Can you update to `0.12`?

```suggestion
  VersionAdded: '0.12'
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### New features
 
 * [#117](https://github.com/rubocop/rubocop-minitest/issues/117): Add new cop `Minitest/AssertWithExpectedArgument` to check for unintended usages of `assert` instead of `assert_equal`. ([@cstyles][])
+* [#124](https://github.com/rubocop/rubocop-minitest/pull/124): Add new `Minitest/AnyAssertions` cop. ([@ghiculescu][])
```
koic, line 10: Can you rebase with the latest mater branch and move this entry to under the `master (unreleased)`?

ghiculescu:
Thanks @koic 

I made all those changes.

lib/rubocop/cop/minitest/any_assertions.rb
```diff
@@ -0,0 +1,48 @@
...
+      #   end
+      #
+      #   # good
+      #   class FooTest < Minitest::Test
+      #     def test_the_truth
+      #       assert true
+      #     end
+      #   end
+      #
+      class AnyAssertions < Cop
```
koic, line 22: Oh, I seem to have forgotten to comment 💦 

I think the cop name `NoAssertionMethods` is clear for registering offenses.

```suggestion
      class NoAssertionMethods < Cop
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
```
koic, line 4: That's just my two cents. Can you insert `### New features` header?

```suggestion
## master (unreleased)

### New features
```

lib/rubocop/cop/minitest/any_assertions.rb
ghiculescu: How about `NoAssertions`, to match `MultipleAssertions` ?

koic: `NoAssertions` also looks good name. And I think `MultipleAssertions` will be different role :-)

ghiculescu: Okay, I can change it tomorrow. But I’m also fine with it now if you want to merge.

koic: Thanks. If you like, can you replace it with new `Base` API for the cop?

```diff
diff --git a/lib/rubocop/cop/minitest/no_assertion_methods.rb b/lib/rubocop/cop/minitest/no_assertion_methods.rb
index 2292d03..36671ee 100644
--- a/lib/rubocop/cop/minitest/no_assertion_methods.rb
+++ b/lib/rubocop/cop/minitest/no_assertion_methods.rb
@@ -19,7 +19,7 @@ module RuboCop
       #     end
       #   end
       #
-      class NoAssertionMethods < Cop
+      class NoAssertionMethods < Base
         include MinitestExplorationHelpers

         MSG = 'Test case has no assertions.'
@@ -32,7 +32,7 @@ module RuboCop

             next if assertions_count.positive?

-            add_offense(node, location: node.block_type? ? :expression : :name)
+            add_offense(node.block_type? ? node.loc.expression : node.loc.name)
           end
         end
```

e.g.: https://github.com/rubocop/rubocop-minitest/pull/128

ghiculescu: @koic updated!

koic:
Great! Thanks @ghiculescu!

andyw8:
Here's another approach to catching tests without assertions:

https://railsatscale.com/2024-01-25-catching-assertionless-tests

PR #138, opened 2021-06-25
Adding new checks for assert_in and refute_in to mirror the assert_includes and refute_includes checks

tayjohno:
So there's already a `assert_includes` and `refute_includes` which each respectively verify that you use `assert_includes` over `assert(collection.includes?(object))`, but there's another method we can check for here, which I think should be helpful. (I've been using a similar check to this at my organization and decided it would be nice to contribute it upstream).

```ruby
# bad
assert(object.in?(collection))
assert(object.in?(collection), 'message')

# good
assert_includes(collection, object)
assert_includes(collection, object, 'message')
```

I recognize that [`Object#in?` is implemented in Rails](https://apidock.com/rails/v6.1.3.1/Object/in%3F), so this cop won't apply to those not using the Rails framework, but it's certainly PRIMARILY a Minitest linter, so I figured this was still the place to implement it.

The implementation of `Object#in?` in Rails is to simply flip the arguments and call `.include?` so these auto-corrects should be safe to change, so long as the `in?` method being matched is truly the Rails `Object#in?`

From apidoc link above:
```
# File activesupport/lib/active_support/core_ext/object/inclusion.rb, line 12
  def in?(another_object)
    another_object.include?(self)
  rescue NoMethodError
    raise ArgumentError.new("The parameter passed to #in? must respond to #include?")
  end
```

-----------------

* [x] Added tests.


koic:
It may be better to extend `Minitest/AssertIncludes` and `Minitest/RefuteIncludes` instead of adding new cops.

- https://github.com/rubocop/rubocop-minitest/blob/master/lib/rubocop/cop/minitest/assert_includes.rb
- https://github.com/rubocop/rubocop-minitest/blob/master/lib/rubocop/cop/minitest/refute_includes.rb

These are suitable as rolls.

koic:
I'm going to close this PR because no activity seen. Anyway it can restart with the following approach:
https://github.com/rubocop/rubocop-minitest/pull/138#issuecomment-874270105

PR #164, opened 2022-03-26
Added new duplicated_test_run cop

ignacio-chiazzo:
**Added new duplicated_test_run cop**

If a Minitest class inherits from another class, it will also inherit its methods causing Minitest to run the parent's tests methods twice.

This PR detects when there are two tests classes, one inherits from the other, and both have tests methods. MinitestRubocop will add an offence to the Child class in such a case. 

For example:
```ruby
class B < Minitest::Test
  test "test 1" do 
      #....
  end

  class A < B
    test "test 2" do 
        #....
    end 
  end
end
```
**It will run `test_1` twice!**

I came across a few PRs/pieces of code using subclassing. [I wrote a Gem](https://github.com/ignacio-chiazzo/ruby-minitest-analyzer) that detects duplicated test run during Runtime and a [Post](https://ignaciochiazzo.medium.com/dont-run-your-ruby-minitest-classes-twice-988645662cdb) with more details.

1_mKcZFLz8LpZHmxI3Qrf_vA

This cop has some limitations:
- It only works if both classes are on the same file. 
- Using `describes` also [has the same problem](https://github.com/ignacio-chiazzo/ruby-minitest-analyzer/issues/11), and the cop doesn't detect it (yet ;)).

**More information [in this post](https://ignaciochiazzo.medium.com/dont-run-your-ruby-minitest-classes-twice-988645662cdb).**


lib/rubocop/cop/minitest/duplicated_test_run.rb
```diff
@@ -0,0 +1,81 @@
...
+      #     end
+      #
+      #     def test_parent
+      #     end
+      #   end
+      #
+      class DuplicatedTestRun < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Subclasses with test methods causes the parent\' tests to run them twice.'
```
koic, line 51:
```suggestion
        MSG = "Subclasses with test methods causes the parent' tests to run them twice."
```

config/default.yml
```diff
@@ -1,3 +1,8 @@
+Minitest/DuplicatedTestRun:
+  Description: 'This cop detects when a test class has test methods, and its parent also has tests methods causing the parents tests to run them twice.'
+  Enabled: pending
+  VersionAdded: '<<next>>'
```
koic, line 4: Can you move it to the position in ASCII order?

CHANGELOG.md
```diff
@@ -286,3 +286,4 @@
 [@cstyles]: https://github.com/cstyles
 [@ghiculescu]: https://github.com/ghiculescu
 [@gi]: https://github.com/gi
+[@ignacio-chiazzo]: https://github.com/ignacio-chiazzo
```
koic, line 289: Can you remove it? Because this line will be auto-generated by release process.

changelog/new_duplicated_test_run_cop.md 
```diff
@@ -0,0 +1 @@
+* [#164](https://github.com/rubocop/rubocop-minitest/pull/164): Add new `Minitest/DuplicatedTestRun cop. ([@ignacio-chiazzo][https://github.com/rubocop/rubocop-minitest/pull/164])
```
koic, line 1:
```suggestion
* [#164](https://github.com/rubocop/rubocop-minitest/pull/164): Add new `Minitest/DuplicatedTestRun cop. ([@ignacio-chiazzo][])
```

config/default.yml
```diff
@@ -1,3 +1,8 @@
+Minitest/DuplicatedTestRun:
+  Description: 'This cop detects when a test class has test methods, and its parent also has tests methods causing the parents tests to run them twice.'
```
andyw8, line 2: This reads a little awkwardly, how about:

```suggestion
  Description: 'This cop detects duplicate test runs caused by one test class inheriting from another.'
```

ignacio-chiazzo: Thanks! I agreed. I updated the description after moving the rule so that it's sorted alphabetically.

config/default.yml
ignacio-chiazzo: I added a test checking that the cops are sorted in the following PR https://github.com/rubocop/rubocop-minitest/pull/165

koic:
Nice! Can you squash your commits into one?

config/default.yml
koic: It may be better to update the cop name to `Minitest/DuplicateTestRuns`. How about it?

koic: @andyw8 Do you have an opinion on this cop name?

andyw8: I'm leaning towards `DuplicateTestRun`. We're looking for duplicate _runs_ across the whole project, but in general we're detecting each duplicate _run_ individually.

ignacio-chiazzo: ok, I updated it to `DuplicateTestRun` and rebased it so that it now checks for sorted cop names in the `default` file. https://github.com/rubocop/rubocop-minitest/pull/165

koic: The name looks good. Thank you!

koic:
Thanks!

PR #165, opened 2022-03-29
Added consitency test for config defaults so that cops are sorted alp…

ignacio-chiazzo:
This PR adds a test to check that the cops are sorted alphabetically.


koic:
I like this improvement 👍 Can you squash commits into one?

ignacio-chiazzo:
>  I like this improvement 👍 Can you squash commits into one?

Thanks, done.

PR #177, opened 2022-07-25
Add new `Minitest/AssertRaisesCompoundBody` cop

### LINKED ISSUE #109: New cop to narrow down `assert_raises` bodies
```ruby
# bad
assert_raises(...) do
  some_other_code
  do_something
end

# good
some_other_code

assert_raises(...) do
  do_something
end
```

It is a good practice to make `assert_raises` bodies as small as possible, so that:
1) it is obvious which method call is tested for error 
2) and when multiple calls inside `assert_raises` can raise the same kind of an exception, there won't be false positives and the right method is tested

So this cop will suggest that `assert_raises` body contains only 1 top level node and it should of `send/csend` type.

@koic Wdyt on this?

fatkodima:
Closes #109.

Tested on a rails codebase: got 51 offense, checked them - seems like is working correctly.


changelog/new_assert_raises_compound_body_cop.md
```diff
@@ -0,0 +1 @@
+* [#177](https://github.com/rubocop/rubocop-minitest/pull/177): Add new `Minitest/AssertRaisesCompoundBody` cop. ([@fatkodima][])
```
koic, line 1:
```suggestion
* [#109](https://github.com/rubocop/rubocop-minitest/issues/109): Add new `Minitest/AssertRaisesCompoundBody` cop. ([@fatkodima][])
```

test/rubocop/cop/minitest/assert_raises_compound_body_test.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class AssertRaisesCompoundBodyTest < Minitest::Test
+  def test_registers_offense_when_multi_statement_bodies
+    assert_offense(<<~RUBY)
+      assert_raises(MyError) do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^ Reduce `#assert_raises` body to contain only the raising code.
```
koic, line 9: Can you tweak the message?

```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^ Reduce `assert_raises` block body to contain only the raising code.
```

fatkodima:
Updated.

koic:
Thanks!

PR #199, opened 2022-11-22
Add new `Minitest/SkipWithoutComment` cop

### LINKED ISSUE #198: New cop to ensure 'skip' has a message
Inspired by - https://hexdocs.pm/credo/Credo.Check.Design.SkipTestWithoutComment.html.

It is possible to `skip` a test without providing a reason why - https://github.com/minitest/minitest/blob/ed224bc9983eab31cbb4859f655daeec903da046/lib/minitest/assertions.rb#L783

Future you or other developers will have a hard time figuring out why it was skipped. So I suggest to enforce `skip` messages. Unless it is in condition already.

Example
```ruby
# bad
skip

# good
skip("Is not supported by Active Record < 7")

# also good
skip if active_record_below_7?
```

fatkodima:
Closes #198.

I am thinking: do we need a config option to be able to always enforce `skip` with comments (no matter if in conditional or not)?


koic:
I think this cop can be published without a config option first. How about a rule like the one below for conditions?

```ruby
# bad
if condition
  do_something
  skip
end

# good
skip if condition

if condition
  skip
end
```

It is clear when the `if` body has a single sentence, but may be ambiguous when there are multiple sentences.

`while` is also ambiguous than `if`, so it may be bad.

```ruby
# bad
do_something while condition
```

And It's a corner case, I think the following is also bad.

```ruby
# bad
if condition
  skip
else
  skip
end
```

Can you add test cases and adjust the implementation?

lib/rubocop/cop/minitest/skip_without_comment.rb
```diff
@@ -0,0 +1,39 @@
...
+      #   skip
+      #   skip('')
+      #
+      #   # good
+      #   skip("Reason why the test was skipped")
+      #
+      #   # good
+      #   skip if condition?
+      #
+      class SkipWithoutComment < Base
```
koic, line 19: How about a name like `SkipWithoutArgument` or `SkipWithoutMessageArgument`? Because "comment" might imagine source code comment.

fatkodima:
@koic Changed the name and adjusted the logic accordingly to the comments.

koic:
I tried running it with the rails/rails repo.

In the following, all `when` not use `skip`, only `else` uses `skip`. In other words, the reason for `skip` in `else` is derivable, so I think it can be allowed. e.g.:

```ruby
          case connection.class::ADAPTER_NAME
          when "PostgreSQL"
            unless connection.instance_variable_get(:@raw_connection).transaction_status == ::PG::PQTRANS_INTRANS
              connection.instance_variable_get(:@raw_connection).async_exec("begin")
            end
            connection.instance_variable_get(:@raw_connection).async_exec("set idle_in_transaction_session_timeout = '10ms'")
            sleep 0.05
          when "Mysql2"
            connection.send(:internal_execute, "set @@wait_timeout=1")
            sleep 1.2
          else
            skip
          end
```

It also seems obvious to `skip` in case of an exception, so it seems that it can be also allowed. e.g.:

```ruby
     begin
       @cache.delete_matched("*")
     rescue NotImplementedError
       skip
     end
```

@fatkodima Can you take a look at this?

fatkodima:
@koic Updated with the suggested cases.

lib/rubocop/cop/minitest/skip_without_reason.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          return if node.parent&.resbody_type?
+
+          add_offense(node)
+        end
+
+        private
+
+        def blank_argument?(node)
+          message = node.arguments.first
```
koic, line 45: That's just my two cents. Can you tweak it?

```suggestion
          message = node.first_argument
```

lib/rubocop/cop/minitest/skip_without_reason.rb
```diff
@@ -0,0 +1,65 @@
...
+          message.nil? || (message.str_type? && message.value == '')
+        end
+
+        def conditional_parent(node)
+          parent = node.parent
+          if parent&.if_type? || parent&.case_type?
+            parent
+          elsif parent&.when_type?
+            parent.parent
+          end
```
koic, line 55: Can you reduce safe navigation operators?

```suggestion
          return unless (parent = node.parent)
          
          if parent.if_type? || parent.case_type?
            parent
          elsif parent.when_type?
            parent.parent
          end
```

test/rubocop/cop/minitest/skip_without_reason_test.rb
```diff
@@ -0,0 +1,114 @@
...
+    assert_no_offenses(<<~RUBY)
+      skip('Reason')
+    RUBY
+  end
+
+  def test_does_not_register_offense_when_skip_with_dynamic_reason
+    assert_no_offenses(<<~RUBY)
+      skip(reason)
+    RUBY
+  end
```
koic, line 30: Can you add a test and update the implement for `obj.skip` that should be allowed?

fatkodima:
Thanks for another review! Added more suggested changes.

koic:
Let's start with this behavior and adjust it according to feedback use cases. Thank you!

PR #217, opened 2022-12-30
Add new cop `Minitest/TestFileName`

### LINKED ISSUE #213: New cop: `Minitest/TestFileName`
This cop should check if we have a test class inside the file without the `_test` suffix. 
In this case, this file will be just ignored when running all the tests and go unnoticed.

tejasbubane:
Closes #213

-----------------

* [x] Added tests.


lib/rubocop/cop/minitest/file_name.rb
```diff
@@ -0,0 +1,39 @@
...
+        def on_class(node)
+          return unless test_class?(node)
+
+          add_offense(node) unless valid_file_name?
+        end
+
+        private
+
+        def valid_file_name?
+          file_name.match?('^test_.*') || file_name.match?('.*_test\.rb$')
```
fatkodima, line 30: `start_with?`/`end_with?` ?
Do we really need to check for `.rb` file extension? 🤔

lib/rubocop/cop/minitest/file_name.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+
+        def file_name
```
fatkodima, line 33: I think we can just inline this into the caller as a local variable.

config/default.yml
```diff
@@ -127,6 +127,11 @@ Minitest/EmptyLineBeforeAssertionMethods:
   Enabled: pending
   VersionAdded: '0.23'
 
+Minitest/FileName:
+  Description: 'Checks if test file names start with `test_` or end with `_test.rb`.'
```
fatkodima, line 131: It would be nice to add a style guide link, when it will be merged - https://github.com/rubocop/rubocop-minitest/issues/213#issuecomment-1368029199

fatkodima (commented):
❤️

lib/rubocop/cop/minitest/file_name.rb
```diff
@@ -0,0 +1,36 @@
...
+      #
+      class FileName < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Test file path should end with `_test.rb`.'
+
+        def on_class(node)
+          return unless test_class?(node)
+
+          add_offense(node) unless valid_file_name?
```
fatkodima, line 24: I think, the offense should be added to the whole file, like in `Lint/EmptyFile` cop.

fatkodima (commented):
A small suggestion, in comment. Overall, lgtm 💪 Thank you!

lib/rubocop/cop/minitest/file_name.rb
```diff
@@ -0,0 +1,36 @@
...
+      #   my_class.rb
+      #
+      #   # good
+      #   my_class_test.rb
+      #   test_my_class.rb
+      #
+      class FileName < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Test file path should end with `_test.rb`.'
```
andyw8, line 19:
```suggestion
        MSG = 'Test file path should start with `test_` or end with `_test.rb`.'
```

lib/rubocop/cop/minitest/file_name.rb
tejasbubane: Updated to use use global message instead.

fatkodima:
Why did you decided to move from `on_class` to `on_new_investigation`? 🤔 Not a big deal, but the previous one looked simpler, imo. 
And I would probably change the name from `FileName` to `TestFileName`, similar to `Minitest/TestMethodName` cop.

tejasbubane:
> Why did you decided to move from on_class to on_new_investigation

@fatkodima Because if a test file has multiple classes, `on_class` will register error for each class.

fatkodima:
Can't we call `add_global_offense` in a `on_class`?

tejasbubane:
> Can't we call add_global_offense in a on_class?

I guess we could, but the cop will be invoked on every class definition which is not needed.

fatkodima:
This shouldn't be a large problem, but yeah, makes sense. Lgtm then 👍

lib/rubocop/cop/minitest/test_file_name.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Checks if test file names start with `test_` or end with `_test.rb`.
```
koic, line 6: I think it can be mentioned that files containing classes with names ending in `Test` are considered test files.

lib/rubocop/cop/minitest/test_file_name.rb
```diff
@@ -0,0 +1,41 @@
...
+      #   my_class_test.rb
+      #   test_my_class.rb
+      #
+      class TestFileName < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Test file path should start with `test_` or end with `_test.rb`.'
+
+        def on_new_investigation
+          return if processed_source.blank?
```
koic, line 22: This logic seems unnecessary, why is it needed?

lib/rubocop/cop/minitest/test_file_name.rb
```diff
@@ -0,0 +1,41 @@
...
+          return unless test_file?(processed_source.ast)
+
+          add_global_offense(MSG) unless valid_file_name?
+        end
+
+        private
+
+        def test_file?(node)
+          test_class?(node) || node.each_child_node.any? { |child_node| test_class?(child_node) }
```
koic, line 31:
```suggestion
          return true if node.class_type? && test_class?(node) 
          
          node.each_descendant(:class).any? { |class_node| test_class?(class_node) }
```

lib/rubocop/cop/minitest/test_file_name.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+
+        def valid_file_name?
+          file_name = File.basename(processed_source.buffer.name)
+          file_name.start_with?('test_') || file_name.end_with?('_test.rb')
```
koic, line 36:
```suggestion
          basename = File.basename(processed_source.file_path)
          
          basename.start_with?('test_') || basename.end_with?('_test.rb')
```

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -23,7 +23,7 @@ module MinitestExplorationHelpers
       private
 
       def test_class?(class_node)
-        class_node.parent_class && class_node.identifier.source.end_with?('Test')
+        class_node&.class_type? && class_node&.parent_class && class_node.identifier.source.end_with?('Test')
```
koic, line 26: Can you revert this change? Because it is assumed that the class node is passed as indicated in the argument name.

test/rubocop/cop/minitest/test_file_name_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  def test_registers_offense_for_invalid_path
+    offenses = inspect_source(<<~RUBY, @cop, 'foo.rb')
+      class FooTest < Minitest::Test
+      end
+    RUBY
+    message = 'Test file path should start with `test_` or end with `_test.rb`.'
+
+    assert_equal(1, offenses.size)
+    assert_equal(message, offenses.first.message)
+  end
```
koic, line 15: It seems that a test to a nested test class like the following is necessary:

```ruby
module Foo
  end
end
```

koic: And Can you add a test for `lib/foo.rb` path?

test/rubocop/cop/minitest/test_file_name_test.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def test_does_not_register_offense_for_files_starting_with_test
+    assert_no_offenses(<<~RUBY, 'test_foo.rb')
+      class FooTest < Minitest::Test; end
+    RUBY
+  end
```
koic, line 21: Ditto. For `'lib/test_foo.rb'`.

test/rubocop/cop/minitest/test_file_name_test.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def test_does_not_register_offense_for_files_ending_with_test
+    assert_no_offenses(<<~RUBY, 'foo_test.rb')
+      class FooTest < Minitest::Test; end
+    RUBY
+  end
```
koic, line 27: Ditto. For `'lib/foo_test.rb'`.

lib/rubocop/cop/minitest/test_file_name.rb
koic: Seeing https://github.com/rubocop/minitest-style-guide/pull/52, I thought it would also mention the risk of tests not running if naming is not followed.

koic:
@tejasbubane ping.

koic:
Thanks!

PR #218, opened 2022-12-30
Add new cop `Minitest/NoTestCases`

### LINKED ISSUE #216: New cop: `Minitest/NoTestCases`
Warns when there are not test cases in the test class. Similar to the existing `NoAssertions` cop.

```ruby
# bad
class SomethingTest < Minitest::Test
  def setup; end
end

# good
class SomethingTest < Minitest::Test
  def setup; end

  def test_something
    assert something
  end
end
```

tejasbubane:
Closes #216

-----------------

* [x] Added tests.


lib/rubocop/cop/minitest/no_test_cases.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      class NoTestCases < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Test class has no test cases.'
+
+        def on_class(node)
+          return unless test_class?(node)
+
+          add_offense(node) unless test_cases(node).count.positive?
```
fatkodima, line 30: Wdyt about `if test_cases(node).empty?`? 
It's interesting, that this was not already suggested by the rubocop itself.

test/rubocop/cop/minitest/no_test_cases_test.rb
```diff
@@ -0,0 +1,32 @@
...
+      class FooTest < Minitest::Test
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Test class has no test cases.
+        def perform
+          1 + 2
+        end
+      end
+    RUBY
+  end
+
+  def test_does_not_register_offense_for_non_test_classes
```
fatkodima, line 25: It would be good to also have a test, that detects no offenses when the test class has a test case. Something like:

```ruby
  def test_foo; end
end
```

fatkodima:
Lgtm 👍

lib/rubocop/cop/minitest/no_test_cases.rb
```diff
@@ -0,0 +1,35 @@
...
+      #   class FooTest < Minitest::Test
+      #     def test_something
+      #       assert true
+      #     end
+      #   end
+      #
+      class NoTestCases < Base
+        include MinitestExplorationHelpers
+
+        MSG = 'Test class has no test cases.'
```
koic, line 25: IMO, I think the action that user should take becomes more explicit:

```suggestion
        MSG = 'Test class should have test cases.'
```

koic:
I tried this cop with [the rails/rails repo](https://github.com/rails/rails) and it seems to find a lot of false positives. Could you take a look?

tejasbubane:
@koic Rails test suite defines lots of temp test classes without any tests. Some examples:

* https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/activerecord/test/cases/validations/uniqueness_validation_test.rb#L49-L60
* https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/railties/test/test_unit/reporter_test.rb#L8-L10
* https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/activesupport/test/cache/stores/redis_cache_store_test.rb#L119-L140

I am not sure how to fix them safely. Any ideas?

koic:
Isn't this rule too strict? I'm concerned that there are more opportunities for false positives than for detected true offenses.　TBH, I don't have a sense of user-friendliness.

fatkodima:
> Rails test suite defines lots of temp test classes without any tests. Some examples:
> 
> https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/activerecord/test/cases/validations/uniqueness_validation_test.rb#L49-L60
> https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/railties/test/test_unit/reporter_test.rb#L8-L10
> https://github.com/rails/rails/blob/b8e09fc54ec369574ff34734514600428dbd8a2a/activesupport/test/cache/stores/redis_cache_store_test.rb#L119-L140
> I am not sure how to fix them safely. Any ideas?

First example: it just incorrectly detects non test class. That should be fixed.
Second example: Some empty test class to test the reporter. Regular projects do not extend test runners and do not have such classes.
Third example: This is a definition of a base test class - a popular practice and in this case the cop should be just disabled inline, in my opinion.

Personally, I see a value of this cop. Maybe not inside the `rails` itself, where "people know what they are doing"™️ and it would just generate false positives (I had not tested on it, probably there would be many valid offenses). But for regular projects this will produce many valid offenses. We can make it disabled by default, for example.

koic:
> We can make it disabled by default, for example.

Yeah, it looks a reasonable idea. @tejasbubane Can you disable this cop by default?

tejasbubane:
@koic Disabled by default.

PR #224, opened 2023-01-15
Fix a false positive for `Minitest/MultipleAssertions` when using assertion method with block

### LINKED ISSUE #223: `Minitest/MultipleAssertions` unexpectedly counts the block node
As far as I see, `Minitest/MultipleAssertions` counts the block node as assertion method, with the [fix](https://github.com/rubocop/rubocop-minitest/commit/55b1800787bef0312451f7b821eacd7c8f8866f0) for #210, so the cop issues a violation unexpectedly if code includes assertion methods that are called with block. I'm not sure how to solve this problem while considering #210, so I just opened an issue here.

## Expected behavior

Given the code below, I expect rubocop-minitest does not issue problems on `Minitest/MultipleAssertions`, which is configured as the default; it means `Max` is `3`.

```ruby
# frozen_string_literal: true

require "test_helper"

class MyMethodTest < Minitest::Test
  def my_method(value)
    raise "error" if value == 2

    value + 2
  end

  test "test!" do
    assert_raises RuntimeError do
      my_method(2)
    end

    assert_equal 3, my_method(1)
    assert_equal 5, my_method(3)
  end
end
```

This is the expected command output:

```
❯ bundle exec rubocop test/my_method_test.rb
Inspecting 1 file
.

1 file inspected, no offenses detected
```

## Actual behavior

```
❯ bundle exec rubocop test/my_method_test.rb
Inspecting 1 file
C

Offenses:

test/my_method_test.rb:12:3: C: Minitest/MultipleAssertions: Test case has too many assertions [4/3].
  test "test!" do ...
  ^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

I created a repository to reproduce this problem.

https://github.com/yykamei/rubocop-minitest-issue

## RuboCop version

```
❯ bundle exec rubocop -V
1.43.0 (using Parser 3.2.0.0, rubocop-ast 1.24.1, running on ruby 3.2.0) [x86_64-darwin21]
  - rubocop-minitest 0.26.0
```

fatkodima:
Fixes #223.

`assertions_count` method was updated to count assertion-like `send` nodes belonging to `block` node only once.


test/rubocop/cop/minitest/multiple_assertions_test.rb
```diff
@@ -15,6 +15,14 @@ def test_asserts_twice
           assert_equal(foo, bar)
           assert_empty(array)
         end
+
+        def test_asserts_three_times
```
koic, line 19: Can you separate this new test example from existing `def test_registers_offense_when_multiple_expectations` (L10) to a new test method?

fatkodima: Done.

koic:
Thanks!

PR #229, opened 2023-01-17
Add new `Minitest/NonPublicTestMethod` cop

### LINKED ISSUE #220: New cop: `Minitest/PrivateTestMethod`
While adding a new test case to a large codebase (`rails/rails` as an example) with its test files sometimes having thousands of lines of code and not following TDD (as I believe most of people do, mine included, except for bugfixes), it is very easy to add a test case to the `private` class section. The test case won't be run and the developer will think that tests are passing.

```ruby
# bad
  def test_foo; end

  private
  def test_bar
  end
end

# good
  def test_foo; end
  def test_bar; end
end

# good
  def test_foo; end

  private
  def bar
  end
end
```

fatkodima:
Closes #220.

Tested on `rails/rails` - no offenses there. 
Does anyone know other large OSS projects with minitest tests in it?


fatkodima:
While testing this cop on https://github.com/openstreetmap/openstreetmap-website, it detected this - https://github.com/openstreetmap/openstreetmap-website/blob/b278bc06ef720af8c21c99659c4d70334384c386/test/integration/oauth2_test.rb#L141, because `rubocop-minitest` detects all class methods starting with `test_` as test cases. But we should ignore such methods having arguments.

koic:
First, it should be enough if it is confirmed with rails/rails repo 👍 Can you update this cop name to `NonPublicTestMethod`? Because "non public" has a naming track record.
https://github.com/rubocop/rubocop/blob/v1.43.0/lib/rubocop/cop/mixin/def_node.rb#L12

config/default.yml
```diff
@@ -158,6 +158,11 @@ Minitest/GlobalExpectations:
   VersionAdded: '0.7'
   VersionChanged: '0.26'
 
+Minitest/HiddenTestMethod:
+  Description: 'Detects hidden (marked as `private` or `protected`) test methods.'
```
koic, line 162:
```suggestion
  Description: 'Detects non public (marked as `private` or `protected`) test methods.'
```

config/default.yml
```diff
@@ -158,6 +158,11 @@ Minitest/GlobalExpectations:
...
+  Enabled: pending
```
koic, line 163: I think it's probably `Severity: warning`.

```suggestion
  Enabled: pending
  Severity: warning
```

lib/rubocop/cop/minitest/hidden_test_method.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Detects hidden (marked as `private` or `protected`) test methods. Minitest runs only
```
koic, line 6:
```suggestion
      # Detects non `public` (marked as `private` or `protected`) test methods. Minitest runs only
```

lib/rubocop/cop/minitest/hidden_test_method.rb
```diff
@@ -0,0 +1,55 @@
...
+      #     def test_does_something
+      #       do_something
+      #     end
+      #   end
+      #
+      class HiddenTestMethod < Base
+        include MinitestExplorationHelpers
+        include DefNode
+
+        MSG = 'Hidden test method detected. Make it public for it to run.'
```
koic, line 45:
```suggestion
        MSG = 'Hidden test method detected. Make it `public` for it to run.'
```

fatkodima:
Updated.

andyw8:
I ran this on Shopify's main app. No errors, and it caught a good number of unintentionally private tests.

fatkodima:
Cool! Glad it works. Thank you for checking ❤️

koic:
Thanks!

PR #230, opened 2023-01-17
Remove duplication from `TestMethodName` cop

fatkodima:
Removed method just duplicates the existing method from `MinitestExplorationHelpers` module.

I have seen similar patterns a few times at least in rubocop itself (for extracting methods; macros; methods and macros; class method nodes; etc from the class nodes), so wondering if its worth extracting it into `rubocop-ast` in the future.


koic:
I haven't delved into the abstraction of the other duplicate code, anyway at least this duplicate can be replaced. Thank you!

PR #231, opened 2023-01-17
Do not treat `test_` methods with arguments as test methods

fatkodima:
While testing https://github.com/rubocop/rubocop-minitest/pull/229 on https://github.com/openstreetmap/openstreetmap-website, it detected this - https://github.com/openstreetmap/openstreetmap-website/blob/b278bc06ef720af8c21c99659c4d70334384c386/test/integration/oauth2_test.rb#L141, because `rubocop-minitest` detects all class methods starting with `test_` as test cases. But we should ignore such methods having arguments.

### DIFF
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -7,6 +7,7 @@ module RuboCop
     # Helper methods for different explorations against test files and test cases.
     # @api private
     module MinitestExplorationHelpers
+      include DefNode
       extend NodePattern::Macros
 
       ASSERTION_PREFIXES = %w[assert refute].freeze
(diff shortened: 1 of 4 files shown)


fatkodima:
Added a test code and changelog entry. I decided to extract test cases into a separate file instead of dumping into some random cop test file. Is it ok?

test/rubocop/cop/mixin/minitest_exploration_helpers_test.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      protected
+      def test_protected; end
+
+      private
+      def test_private; end
+    end
+  RUBY
+
+  def test_test_case_p_returns_true_for_test_case
```
andyw8, line 27: What does the `p` mean here?

fatkodima: "predicate". It's a replacement for `?` at the end of the method name. I saw this convention in the ruby source code. For example, https://github.com/ruby/ruby/blob/a9bcc058bb4cd954e8406c93debb876983627ca9/array.c#L8851.

test/rubocop/cop/mixin/minitest_exploration_helpers_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def test_test_case_p_returns_false_for_test_methods_with_arguments
+    refute Helper.test_case?(method_node(SOURCE, :test_with_arguments))
+  end
+
+  private
+
+  def method_node(source, method_name)
+    class_node = parse_source!(source).ast
```
koic, line 47: It could be reduced to the single argument. The entire place where `SOURCE` is used.

```suggestion
  def test_test_case_p_returns_false_for_test_methods_with_arguments
    refute Helper.test_case?(method_node(:test_with_arguments))
  end

  private

  def method_node(method_name)
    class_node = parse_source!(SOURCE).ast
```

test/rubocop/cop/mixin/minitest_exploration_helpers_test.rb
koic: Even if `p` is used in the product code, the `test_` descriptions might be easier to understand with `predicate`.
On a side note, `p` reminds me of LISP :-)

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -43,6 +44,10 @@ def test_cases(class_node)
         test_cases + test_blocks
       end
 
+      def test_case_signature?(def_node)
```
koic, line 47: `test_method?` might be better, AFAIK signature doesn't include scope.

```suggestion
      def test_method?(def_node)
```

fatkodima:
Updated with the suggestions.

koic:
Thanks!

PR #232, opened 2023-01-19
Add new `Minitest/LifecycleHooksOrder` cop

### LINKED ISSUE #214: New cop: `Minitest/HooksGrouping`
Or there is a better name for this cop. 
It will check that `setup`/`teardown` are *before* all the test cases, and that `setup` is before `teardown`. It adds consistency and enforces the expectation of these methods to go first (to be easily noticeable/etc).

```ruby
# bad
class SomethingTest < Minitest::Test
  def setup; end
  def test_something; end
  def teardown; end
end

# bad
class SomethingTest < Minitest::Test
  def teardown; end
  def setup; end
  def test_something; end
end

# good
class SomethingTest < Minitest::Test
  def setup; end
  def teardown; end
  def test_something; end
end
```

fatkodima:
Closes #214.

Tested on `rails/rails` - 5 valid offenses.


rubocop-minitest.gemspec
```diff
@@ -32,6 +32,6 @@ Gem::Specification.new do |spec|
   spec.executables = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.add_runtime_dependency 'rubocop', '>= 0.90', '< 2.0'
+  spec.add_runtime_dependency 'rubocop', '>= 1.43', '< 2.0'
```
fatkodima, line 35: Upgraded rubocop because I need `range_with_comments_and_lines` from `RangeHelp` module. 
Probably better done as a separate PR?

lib/rubocop/cop/minitest/lifecycle_hooks_order.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      class LifecycleHooksOrder < Base
+        include MinitestExplorationHelpers
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = '`%<current>s` is supposed to appear before `%<previous>s`.'
+
+        HOOKS_ORDER_MAP = Hash.new do |hash, hook|
+          hash[hook] = LIFECYCLE_HOOK_METHODS_IN_ORDER.index(hook) || 100
```
koic, line 54: Can you make the magic number `100` a constant and add an explanation of what it is?

rubocop-minitest.gemspec
koic: It's a big jump, but let's still include it in this PR. By the way, it looks like this was introduced in 1.39, not 1.43.
https://github.com/rubocop/rubocop/commit/0d6b0d126f5ce887b496bd4ab801e53fad0c8956

```suggestion
  spec.add_runtime_dependency 'rubocop', '>= 1.39', '< 2.0'
```

fatkodima:
Updated.

lib/rubocop/cop/minitest/lifecycle_hooks_order.rb
```diff
@@ -0,0 +1,99 @@
...
+      #   end
+      #
+      class LifecycleHooksOrder < Base
+        include MinitestExplorationHelpers
+        include RangeHelp
+        extend AutoCorrector
+
+        MSG = '`%<current>s` is supposed to appear before `%<previous>s`.'
+
+        REGULAR_METHOD_POSITION = 100
```
koic, line 53: It would be better if source commented why the value of `100` was chosen and set :-)

fatkodima: Updated.

koic:
Thanks!

PR #237, opened 2023-01-29
Fix a false positive for `Minitest/UselessAssertion` when using command execution

fatkodima:
Found a false positive: 
```ruby
assert_empty(`bundle exec yard --no-save --no-output --no-stats`)
```
https://github.com/fatkodima/sidekiq-iteration/blob/81453f8a5586367ba69c80aeea960fe824b76c40/test/documentation_test.rb#L8


koic:
Thanks!

PR #240, opened 2023-02-11
Enhance `AssertMatch`/`RefuteMatch` to check `match?` and `=~` methods

### LINKED ISSUE #239: `Minitest/AssertMatch` behavior doesn't match style guide
Currently, [the style guide](https://minitest.rubystyle.guide/#assert-match) for `Minitest/AssertMatch` says:

```ruby
# bad
assert(pattern.match?(object))
# good
assert_match(pattern, object)
```

However, the cop detects a different pattern:

```ruby
assert(pattern.match(object))
```

Ideally, it should detect all of these variants:
```ruby
assert(pattern.match(object))
assert(pattern.match?(object))
assert(object =~ pattern)
```

fatkodima:
Closes #239.


lib/rubocop/cop/mixin/minitest_cop_rule.rb
```diff
@@ -24,7 +24,8 @@ module MinitestCopRule
       # @api private
       #
       def define_rule(assertion_method, target_method:, preferred_method: nil, inverse: false)
-        preferred_method = "#{assertion_method}_#{target_method.to_s.delete('?')}" if preferred_method.nil?
+        target_methods = Array(target_method)
+        preferred_method = "#{assertion_method}_#{target_methods.first.to_s.delete('?')}" if preferred_method.nil?
```
koic, line 28: It can be confusing to see the special treatment of the first element. So, the order of the array shouldn't matter for API user.

Perhaps it would be preferable to require `preferred_method` optional param if more than one element is specified. It might be helpful to add it to the documentation and throw an exception message if it was not specified the option.

fatkodima: Thanks! Added a check and updated the docs.

lib/rubocop/cop/mixin/minitest_cop_rule.rb
```diff
@@ -24,7 +29,12 @@ module MinitestCopRule
       # @api private
       #
       def define_rule(assertion_method, target_method:, preferred_method: nil, inverse: false)
-        preferred_method = "#{assertion_method}_#{target_method.to_s.delete('?')}" if preferred_method.nil?
+        target_methods = Array(target_method)
+        if target_methods.size > 1 && preferred_method.nil?
+          raise ArgumentError, 'preferred_method should be provided if using more than one target methods'
```
koic, line 34: Can you tweak it?

```suggestion
          raise ArgumentError, '`:preferred_method` keyword argument must be used if using more than one target methods.'
```

fatkodima: Done.

koic:
Thanks!

PR #249, opened 2023-04-16
Handle assertions in conditionals branches in `Minitest/MultipleAssertions` cop

fatkodima:
These changes are to allow to have assertions like:
```ruby
def test_foo
  if ActiveRecord::VERSION >= "7.0"
    assert ActiveRecord.old_foo_method(1, 2)
  else
    assert ActiveRecord.new_foo_method(1, 2)
  end 
end
```

A recent example that inspired this - https://github.com/ixti/sidekiq-throttled/pull/142/files#diff-87c291ea2af263111982285d3dea3083faa62260cc6080cc5396f4bf82a37cf4R22-R27 (it is using rspec, but it does not matter)

### DIFF
--- a/lib/rubocop/cop/minitest/multiple_assertions.rb
+++ b/lib/rubocop/cop/minitest/multiple_assertions.rb
@@ -3,7 +3,8 @@
 module RuboCop
   module Cop
     module Minitest
-      # Checks if test cases contain too many assertion calls.
+      # Checks if test cases contain too many assertion calls. If conditional code with assertions
+      # is used, the branch with maximum assertions is counted.
       # The maximum allowed assertion calls is configurable.
       #
       # @example Max: 1
@@ -36,7 +37,7 @@ module RuboCop
           return unless test_class?(class_node)
 
           test_cases(class_node).each do |node|
-            assertions_count = assertions_count(node)
+            assertions_count = assertions_count(node.body)
 
             next unless assertions_count > max_assertions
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/minitest/multiple_assertions.rb
```diff
@@ -24,6 +24,14 @@ module Minitest
...
       #       assert_empty(array)
       #     end
+      #
+      #     def test_conditional_assertion
+      #       if condition
+      #         assert foo
+      #       else
+      #         assert bar
+      #       end
+      #     end
```
koic, line 34: I worry about losing the symmetry between `#bad` and `#good` examples. Can you replace the added example code with a description of the cop using natural language?

lib/rubocop/cop/minitest/multiple_assertions.rb
```diff
@@ -49,6 +57,29 @@ def on_class(class_node)
 
         private
 
+        def assertions_count(node)
+          return 0 unless node.is_a?(RuboCop::AST::Node)
+
+          assertions =
+            case node.type
+            when :if, :case
```
koic, line 65: Same for `case_match`?

lib/rubocop/cop/minitest/multiple_assertions.rb
```diff
@@ -49,6 +57,29 @@ def on_class(class_node)
...
+              assertions_count_in_branches(node.branches)
+            when :rescue
+              assertions_count(node.body) + assertions_count_in_branches(node.branches)
+            when :block
```
koic, line 69: Same for `numblock`?

test/rubocop/cop/minitest/multiple_assertions_test.rb
```diff
@@ -93,26 +93,193 @@ def test_generates_a_todo_based_on_the_worst_violation
...
+                assert foo
+                assert foo
+              else
+                assert foo
+                assert foo
+              end
+            else
+              assert foo
+            end
+          end
```
koic, line 251: Can you add a source code comment to the line which line is being counted?

test/rubocop/cop/minitest/multiple_assertions_test.rb
```diff
@@ -93,26 +93,193 @@ def test_generates_a_todo_based_on_the_worst_violation
...
+              case
+              when condition
+                assert foo
+              else
+                assert foo
+              end
+            else
+              assert foo
+            end
+          end
```
koic, line 282: Ditto.

fatkodima:
Applied suggested changes. CI is failing, because rubocop now requires ruby 2.7, but this gem - ruby 2.6. So it is needed to first drop support for ruby 2.6 and then this PR can be updated.

Should I open a separate PR with that change or should you?

koic:
> Should I open a separate PR with that change or should you?

The build error has been fixed by https://github.com/rubocop/rubocop-minitest/pull/250. Either way, I understand that the issue is unrelated to this PR change. Thank you :-)

fatkodima:
Updated, now CI is green 🟢

PR #251, opened 2023-04-22
Add new `Minitest/ReturnInTestCase` cop

### LINKED ISSUE #236: New cop to detect `return`s in test cases
```ruby
# bad
def test_check_constraints
  return if mysql?
  # do testing
end

# good
def test_check_constraints
  skip("MySQL does not support check constraints") if mysql?
  # do testing
end
```
Example: https://github.com/gregnavis/active_record_doctor/blob/fc736812db046e2c0bfdbb6cec0b837e01a35076/test/active_record_doctor/detectors/mismatched_foreign_key_type_test.rb#L6

When using `skip`, minitest will report a skipped tests summary at the end; `--verbose` and `--show-skips` will show skips (https://github.com/minitest/minitest/blob/0984e29995a5c0f4dcf3c185442bcb4f493ed5e3/lib/minitest.rb#L208-L214); [minitest-reporters](https://github.com/minitest-reporters/minitest-reporters) will show the skipped tests; `skip` requires a reason with the existing cop (`SkipWithoutReason`) and overall it is more clear the intention, than just simple `return`.

fatkodima:
Closes #236.

Tested on `rails/rails` - 11 valid offenses.


lib/rubocop/cop/minitest/return_in_test_case.rb
```diff
@@ -0,0 +1,38 @@
...
+      #
+      class ReturnInTestCase < Base
+        include MinitestExplorationHelpers
+        extend AutoCorrector
+
+        MSG = 'Use `skip` instead of `return`.'
+
+        def on_return(node)
+          return unless node.ancestors.any? { |parent| test_case?(parent) }
+          return if node.ancestors.find(&:block_type?)
```
koic, line 29: Is `numblock` the same?

fatkodima: Thanks! Always forgetting this ruby feature.

koic: Thanks too :-) Can you add a test for that?

fatkodima: Already added 😉

koic: Oops! I was overlooking. Thank you again!

koic:
@fatkodima I noticed it before release. Considering offense detection target, `Minitest/ReturnInTestMethod` may be better cop name than `Minitest/ReturnInTestCase`. WDYT?

fatkodima:
@koic I am fine with both names. Feel free to change the name or I can change, just let me know.

koic:
Thank you. I opened https://github.com/rubocop/rubocop-minitest/pull/253.

PR #264, opened 2023-09-26
Update AssertOperator and RefuteOperator to allow :[]

flavorjones:
Update the `AssertOperator` and `RefuteOperator` cops to allow `:[]`

The rubocop style guide prefers `assert` and `refute` for testing booleans. If I have an array of booleans, then I want to be able to use this style:

```
  def test_my_array
    assert(array_of_bools[0])
    refute(array_of_bools[1])
  end
```

This PR is suggested by an [actual test in the Nokogiri suite](https://github.com/sparklemotion/nokogiri/blob/main/test/xml/test_reader.rb#L699-L700) which collects boolean values from Nokogiri::Reader as it passes through the document, and then checks the values after completing the parsing pass. That test could be rewritten, but I don't personally think it would be made more readable by that rewrite.

Also see https://github.com/rubocop/rubocop-minitest/pull/260#issuecomment-1734520785 which raised the question of whether `:[]` was intentionally a violation.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/minitest/assert_operator.rb
+++ b/lib/rubocop/cop/minitest/assert_operator.rb
@@ -18,10 +18,14 @@ module RuboCop
 
         MSG = 'Prefer using `assert_operator(%<new_arguments>s)`.'
         RESTRICT_ON_SEND = %i[assert].freeze
+        ALLOWED_OPERATORS = [:[]].freeze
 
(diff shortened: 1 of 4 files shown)


koic:
This seems like a reasonable change. Thank you!

PR #272, opened 2023-10-29
Add new `Minitest/RedundantMessageArgument` cop

koic:
This PR adds new `Minitest/RedundantMessageArgument` cop that detects redundant message argument in assertion methods. The message argument `nil` is redundant because it is the default value.

```ruby
# bad
assert_equal(expected, actual, nil)

# good
assert_equal(expected, actual)
assert_equal(expected, actual, 'message')
```

-----------------

* [x] Added tests.


andyw8:
I'm curious about the context for this, where did you encounter it?

koic:
https://github.com/zdennis/activerecord-import/commit/85cbdea5cc2a759c5224890cef5aada29651646b

andyw8:
I see. I'm wondering if it might be better for this to be in Minitest, e.g. it can raise if `nil` is given for that argument?

koic:
That's an interesting idea. If you're up for it, please propose it to Minitest. 
Regardless, `raise` operates at runtime, while this cop is for static analysis, so the detection timing differs. It shouldn't detract from the value of this cop, I think.

PR #276, opened 2023-11-25
Enhance `AssertSame`/`RefuteSame` to check for `object_id` comparison

fatkodima:
It is a very popular pattern:
```ruby
# bad
assert_equal(expected.object_id, actual.object_id)

# good
assert_same(expected, actual)
```

Tested on `rails/rails` codebase and got 11 offenses.

### DIFF
--- a/changelog/change_assert_same_check_object_ids.md
+++ b/changelog/change_assert_same_check_object_ids.md
@@ -1,0 +1,1 @@
+* [#276](https://github.com/rubocop/rubocop-minitest/pull/276): Enhance `AssertSame`/`RefuteSame` to check for `object_id` comparison. ([@fatkodima][])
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/mixin/assert_refute_same_helper.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Provide a method to define offense rule for Minitest cops.
+    module AssertRefuteSameHelper
+      def define_rule(assertion_method)
+        class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
```
koic, line 8: Since it's specialized for `Minitest/AssertSame` and `Minitest/RefuteSame`, there's likely no need for metaprogramming with eval.

fatkodima: But we define different things depending on where we use this method.

fatkodima: Updated.

koic:
Thanks!

PR #278, opened 2023-12-05
Make the RuboCop::Minitest::Test suite re-usable by gem consumers

DougEdey:
I've been writing custom cops for various jobs for a while now, but I've never had a consistent way to write tests for products that use minitest as the test suite (officially Rubocop only provides `rspec` helpers) , I started to look around to see what this Gem did and figured it would make sense to make `rubocop-minitest` provide a minitest class for others that write custom cops for their own code bases.

This does a couple of things

1. Makes a new `Rubocop::Minitest::Test` class that includes the code from `assert_offense` (I don't feel like it needs to be a module), and it also uses declarative testing for people to use.
2. I moved the "real" test suites in this gem to use the new base class, and kept the code blocks being tested as `Minitest::Test`
3. I added [declarative testing](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Declarative.html) since all the test suites I've been in have it, where instead of using `def test_something_does_something...end` you can do `test "something does something" do...end` which makes it easier to read test names.

I tested this locally in an app of mine and it works great! So it can allow a consistent way for people to write custom cops and easily test them!

-----------------

* [-] Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* [x] Added tests.


exterm:
This makes a lot of sense to me and I'd love to use it.

Earlopain:
If this is being exposed like this then I think it needs some documetation on how to pass global and cop specific config to a test.

See https://github.com/rubocop/rubocop-minitest/blob/b446022ea09b3f5558df9c0106c9e714c6fc1ec5/test/rubocop/cop/minitest/global_expectations_test.rb#L504-L513 and https://github.com/rubocop/rubocop-minitest/blob/b446022ea09b3f5558df9c0106c9e714c6fc1ec5/test/rubocop/cop/minitest/multiple_assertions_test.rb#L339-L343

lib/rubocop/minitest/support.rb
```diff
@@ -1,10 +1,259 @@
...
 
 require 'rubocop'
 require 'minitest/autorun'
 require 'minitest/pride'
-require_relative 'assert_offense'
+require 'rubocop/rspec/expect_offense'
+require 'rubocop/cop/legacy/corrector'
 
-Minitest::Test.include RuboCop::Minitest::AssertOffense
+module Rubocop
```
Earlopain, line 12: Usual capitalization for this is `RuboCop`

lib/rubocop/minitest/support.rb
```diff
@@ -1,10 +1,259 @@
 # frozen_string_literal: true
 
 # Require this file to load code that supports testing using Minitest.
+# rubocop:disable Metrics/ClassLength
 
 require 'rubocop'
 require 'minitest/autorun'
 require 'minitest/pride'
```
Earlopain, line 8: This should probably go when others are using this as well

lib/rubocop/minitest/support.rb
```diff
@@ -1,10 +1,259 @@
...
+        expected_annotations = RuboCop::RSpec::ExpectOffense::AnnotatedSource.parse(source)
+        actual_annotations = expected_annotations.with_offense_annotations(offenses)
+
+        assert_equal(source, actual_annotations.to_s)
+      end
+
+      def assert_offense(source, file = nil, **replacements)
+        setup_assertion
+
+        @cop.instance_variable_get(:@options)[:autocorrect] = true
```
Earlopain, line 144: If the testfile isn't named properly this results in a `NoMethodException`

lib/rubocop/minitest/support.rb
DougEdey: I don't follow, do you want me to remove this line entirely?

lib/rubocop/minitest/support.rb
DougEdey: Fixed, I made `setup_assertion` raise if there's no `@cop` defined.

DougEdey:
@Earlopain Updated the casing of `RuboCop`, updated the sample docs (both inline and `README.md`), handling the nil case for `@cop`

lib/rubocop/minitest/support.rb
Earlopain: If I decide to use this all my logs are suddenly absolutely fabulous since it includes the minitest pride plugin

DougEdey: Ok, I think I understand a bit better, I've moved it to the `test_helper`

lib/rubocop/minitest/support.rb
```diff
@@ -1,10 +1,269 @@
...
+    #   end
+    class Test < ::Minitest::Test
+      class << self
+        # Helper to define a test method using a String. Under the hood, it replaces
+        # spaces with underscores and defines the test method.
+        #
+        #   test "verify something" do
+        #     ...
+        #   end
+        def test(name, &block)
```
Earlopain, line 96: Where is this method coming from? I don't think it was here before, but if it was it seems out of scope to be part of this class

DougEdey: This is a new test method, I updated the PR description (I forgot to add it), it introduces [declarative testing](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Declarative.html) since all the apps I've worked in have it in some form.

Instead of always writing

```ruby
def test_something_does_something
  ...
end
```

You can do

```ruby
test "something does something" do
end
```

Which in my experience is easier to read, and you can avoid confusion with `def test_my_method_with_underscores_in_the_name` and `test "my_method with underscores in the name"` which improves test readability to me

DougEdey: It doesn't replace the `def test_...` definitions but works along side

Earlopain: There are gems for that purpose, `minitest/spec` itself comes pretty close with its [it](https://docs.seattlerb.org/minitest/Minitest/Spec/DSL.html#method-i-it) method. If I want to use this test method I shouldn't also get methods to assert RuboCop cops. It would only be available when in that context anyways.

DougEdey: OK removed it /shrug

DougEdey:
Squashed commits and confirmed CI is passing locally

DougEdey:
@koic 👋 Could I ask for a review?

DougEdey:
@koic  it's been a while, but I'd still like to get this merged

Earlopain:
It's been a while since I last looked at this and this may be a dumb question to only come now but can you explain why this is needed?

Specifically I think there already exists what you are looking for: https://github.com/rubocop/rubocop-minitest/issues/155 and https://github.com/rubocop/rubocop-minitest/pull/156

https://github.com/rubocop/rubocop-minitest/blob/master/lib/rubocop/minitest/support.rb
There may be concern about it polluting the main test class but even then you can just make your own and mix that module in yourself.

koic:
The revised API name and implementation have been provided in #355, so I'll close this one. Thanks for the proposal.

PR #294, opened 2024-01-02
Raise an exception if tests or changelog generation fails

G-Rath:
I realised while working on my recent PRs that CI wasn't actually failing when `bundle exec rake [test]` reported errors - turns out that tests are being invoked by `system` which by default does not do any error handling; instead it sets `$?` to the exit code which can then be checked.

Ruby 2.6 added an `exception` option which when `true` causes `system` to raise an exception if the command exits with a non-zero code, making this very easy to address.

I've checked the codebase and confirmed these are the only uses of `system`.

-----------------

* [x] ~Added tests.~

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,7 +21,7 @@ require_relative 'lib/rubocop/cop/generator'
 
 desc 'Run tests'
 task :test do
-  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}")
+  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}", exception: true)
 end
 
 desc 'Run RuboCop over itself'
--- a/Rakefile
+++ b/Rakefile
@@ -21,7 +21,9 @@ require_relative 'lib/rubocop/cop/generator'
 
 desc 'Run tests'
 task :test do
-  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}", exception: true)
+  error_on_failure = ENV.fetch('ERROR_ON_TEST_FAILURE', 'true') != 'false'
+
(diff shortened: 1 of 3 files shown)


G-Rath:
Ok so it looks like the JRuby and "oldest supported RuboCop version" jobs are already failing which this change has revealed - I've added an env variable to control if test failures should actually be surfaced to use in those two jobs so that this PR can be landed as a dedicated change, and then those jobs can be addressed in their own dedicated PRs (since "mostly working CI" is better than "not working CI" and I have no context on either of the failures so expect they'll take more time to resolve).

I've kept this as a separate commit in case you want to do something different @koic 

Also it looks like `MultipleAssertionsTest#test_generates_a_todo_based_on_the_worst_violation` is flakey - I've confirmed this locally: running multiple times, every so often I get a failure with an incorrect todo count; again something to resolve in a dedicated PR - for now just rerunning that job for this PR should get it passing.

Rakefile
```diff
@@ -21,7 +21,9 @@ require_relative 'lib/rubocop/cop/generator'
 
 desc 'Run tests'
 task :test do
-  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}")
+  error_on_failure = ENV.fetch('ERROR_ON_TEST_FAILURE', 'true') != 'false'
+
+  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}", exception: error_on_failure)
```
koic, line 26: Is there any problem with always setting it to `true` for `exception` option?

```suggestion
  system("bundle exec minitest-queue #{Dir.glob('test/**/*_test.rb').shelljoin}", exception: true)
```

G-Rath: Yes, two of the jobs are currently failing and I think fixing them should be done in dedicated PRs - once they're fixed, this can then be set to `true`.

koic: Ah, OK. Let's merge this first 👍

PR #296, opened 2024-01-03
Don't use `test-queue` to run tests on JRuby because it doesn't support `fork`

G-Rath:
Follow up to #294

Currently this is aimed at being the most straightforward if probably slightly inelegant way to get tests running on JRuby - I assume there is probably a slightly cleaner way of invoking tests but I don't yet know what it is and `Rake::TestTask` was more than a few lines of code 😅

I think it would be worth considering having `test-queue` "support" JRuby by not using `fork` if that platform is detected or if you set `TEST_QUEUE_WORKERS` to 1 (or maybe 0?) as a way of making it easier to have a normalized way of invoking tests across different platforms like this; but it would also be equally valid I think for `test-queue` to say that's out of scope for the gem🤷 

-----------------

* [x] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,11 +19,20 @@ require 'rubocop/rake_task'
 require 'rake/testtask'
 require_relative 'lib/rubocop/cop/generator'
 
-desc 'Run tests'
-task :test do
-  error_on_failure = ENV.fetch('ERROR_ON_TEST_FAILURE', 'true') != 'false'
+# JRuby and Windows don't support fork, so we can't use minitest-queue.
+if RUBY_ENGINE == 'jruby' || RuboCop::Platform.windows?
+  Rake::TestTask.new do |t|
+    t.libs << 'test'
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -19,11 +19,20 @@ require 'rubocop/rake_task'
 require 'rake/testtask'
 require_relative 'lib/rubocop/cop/generator'
 
-desc 'Run tests'
-task :test do
-  error_on_failure = ENV.fetch('ERROR_ON_TEST_FAILURE', 'true') != 'false'
+# jruby does not support fork, so we can't use minitest-queue
+if RUBY_PLATFORM == 'java'
```
koic, line 23: Can you tweak it?

```suggestion
# JRuby and Windows don't support fork, so we can't use minitest-queue.
if RUBY_ENGINE == 'jruby' || RuboCop::Platform.windows?
```

koic:
Thanks!

PR #302, opened 2024-02-14
[Fix #301] Add new Minitest/Focus cop

### LINKED ISSUE #301: New cop suggestion: detect focussed tests

[minitest-focus](https://github.com/minitest/minitest-focus) is a great utility to focus in on one test at a time. But the risk is that someone accidentally leaves a test focussed while committing, and nobody notices the test suite is now only one test long.

I'd like to see a test that catches this, and fails if any focussed tests are found.

## Additional context

rubocop-rspec has the equivalent cop to this - see https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecfocus

jaredmoody:
Checks for focused tests in the formats supported by [minitest-focus](https://github.com/minitest/minitest-focus).

I decided not to add an auto-corrector because IME it gets in the way when using editor auto-formatting. When focusing a test and then saving the file while working on the focused test, the focus gets removed, running everything.

However, I could add this if others disagree.

Hat tip to @NewAlexandria for the [original PR.](https://github.com/rubocop/rubocop/pull/12399/files)

-----------------

* [x] Added tests.


lib/rubocop/cop/minitest/no_focus.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Enforces tests are not focused
```
koic, line 6:
```suggestion
      # Checks if tests are focused.
```

lib/rubocop/cop/minitest/no_focus.rb
```diff
@@ -0,0 +1,34 @@
...
+      #   # bad
+      #   focus
+      #   test 'foo' do
+      #   end
+      #
+      #   # good
+      #   test 'foo' do
+      #   end
+      #
+      class NoFocus < Base
```
koic, line 18: Can you rename to `Minitest/Focus` similar to `RSpec/Focus`?

```suggestion
      class Focus < Base
```

koic:
The `focus` method that is not modifying the `test` method should not trigger warnings (e.g., calls to `focus` method alone). Can you add tests for this case and update the implementation accordingly?

koic:
> I decided not to add an auto-corrector because IME it gets in the way when using editor auto-formatting. When focusing a test and then saving the file while working on the focused test, the focus gets removed, running everything.

While it has not been released yet, the next release of RuboCop will enable control through `--editor-mode`. Cops set with `AutoCorrect: contextual` will not autocorrect when `--editor-mode` is used.
https://github.com/rubocop/rubocop/pull/12682

jaredmoody:
> The focus method that is not modifying the test method should not trigger warnings (e.g., calls to focus method alone). Can you add tests for this case and update the implementation accordingly?

Per the [README](https://github.com/minitest/minitest-focus), minitest-focus allows focusing by directly - by placing the focus on the same line, or indirectly, on the previous line, which sets a "trap", focusing the next test. I added an additional "bad" example showing both focus possibilities.

> While it has not been released yet, the next release of RuboCop will enable control through --editor-mode. Cops set with AutoCorrect: contextual will not autocorrect when --editor-mode is used.

This is cool! I think an autocorrect for this is lower value, since usually only a single test will be focused, but I'll take a stab at autocorrecting. Is there anything I need to do to set this cop to AutoCorrect: contextual?

jaredmoody:
I've added auto-correction and tests are passing, but as I've never written one of these before, I don't know if I did it the best way, suggestions welcome.

koic:
> Per the [README](https://github.com/minitest/minitest-focus), minitest-focus allows focusing by directly - by placing the focus on the same line, or indirectly, on the previous line, which sets a "trap", focusing the next test. I added an additional "bad" example showing both focus possibilities.

Oops! I was mistaken 🙇  You're correct!

> Is there anything I need to do to set this cop to AutoCorrect: contextual?

Since `AutoCorrect: contextual` has not been released yet, it's fine to keep the current behavior of autocorrection as is. It can be addressed later once supported.

lib/rubocop/cop/minitest/focus.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Minitest
+      # Enforces tests are not focused
```
koic, line 6: Please add a period at the end of the sentence.

```suggestion
      # Enforces tests are not focused.
```

lib/rubocop/cop/minitest/focus.rb
```diff
@@ -0,0 +1,47 @@
...
+          return unless focused?(node)
+
+          add_offense(node) do |corrector|
+            if node.child_nodes.empty?
+              corrector.remove(range_by_whole_lines(node.source_range, include_final_newline: true))
+            else
+              corrector.replace(node, node.child_nodes.map(&:source).join(' '))
+            end
+          end
+        end
```
koic, line 43: Can you update to the following? 

```suggestion
        def on_send(node)
          return unless node.receiver

          range = node.loc.selector

          add_offense(range) do |corrector|
            corrector.remove(range)
          end
        end
```

This change will make the offense highlight range clearer. While the layout may be slightly off after autocorrection, let's leave it to `Layout` cops to handle. So, Please adjust the test code accordingly.

test/rubocop/cop/minitest/focus_test.rb
```diff
@@ -0,0 +1,49 @@
...
+
+require_relative '../../../test_helper'
+
+class FocusTest < Minitest::Test
+  def test_registers_offense_when_using_direct_focus
+    assert_offense(<<~RUBY)
+      class FooTest < Minitest::Test
+        focus test 'foo' do
+        ^^^^^^^^^^^^^^^^ Remove `focus` from tests.
+        end
```
koic, line 11: The logic above can clarify the target:

```suggestion
        focus test 'foo' do
        ^^^^^ Remove `focus` from tests.
        end
```

koic:
Thanks!

PR #308, opened 2024-04-07
Make `UselssAssertion` autocorrect `contextual`

Earlopain:
This is an annoying autocorrect to happen while typing.

-----------------

* [ ] Squashed related commits together.
* [ ] Added tests.

### DIFF
--- a/changelog/change_make_useless_assertion_as_contextual.md
+++ b/changelog/change_make_useless_assertion_as_contextual.md
@@ -1,0 +1,1 @@
+* [#308](https://github.com/rubocop/rubocop-minitest/pull/308): Make `UselssAssertion` autocorrect `contextual`. ([@earlopain][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -349,3 +349,4 @@ Minitest/UselessAssertion:
   Description: 'Detects useless assertions (assertions that either always pass or always fail).'
   Enabled: pending
   VersionAdded: '0.26'
+  AutoCorrect: contextual


koic:
Hm, I don't see the use case for setting this cop to `AutoCorrect: contextual`.

Earlopain:
Here's how it goes for me usually

[Screencast_20240407_173508.webm](https://github.com/rubocop/rubocop-minitest/assets/14981592/afcd2022-7cf2-47c0-880b-6cc9af068500)

koic:
Thank you for making the video. However, I didn't quite get it. Can you provide some additional explanation in text?

Earlopain:
Sure, I can try. When I write tests I do this tend to do this:

```
assert_equal([], some_method)
# Expected: []
# Actual: ["foo", "bar"]
```

I let the tests run, get the data from the failure, and copy it over to the first argument to make it pass. With the autocorrect I have an extra step where I need to restore the intended `assert_equal` to make it work.

Hope that makes sense.

andyw8:
I understand the annoyance for your workflow, but I think this shouldn't be marked contextual. I think there would be many genuine occurrences that won't be caught until a full RuboCop check is run, e.g. after pushing to CI.

koic:
I agree with @andyw8. So, applying `AutoCorrect: contextual` for cases clearly written with unintended arguments would mean that most cops in RuboCop Minitest would no longer be `AutoCorrect: always` (default).
I will close this PR, but thank you for the video and the explanation.

PR #314, opened 2024-08-12
Tests using AssertOffense should Fail with a usable message if the relevant cop is not defined

brandoncc:
Including the `AssertOffense` module currently fails with a cryptic message if the cop can't be found. This PR introduces an explicit failure with a useful message, replacing the `NoMethodError` that is currently raised in this situation.

**Before**

```
Error:
RuboCop::Cop::MyCops::MyNewCopTest#test_works:
NoMethodError: undefined method `[]=' for nil
    /Users/brandoncc/.gem/ruby/3.3.1/gems/rubocop-minitest-0.35.1/lib/rubocop/minitest/assert_offense.rb:109:in `assert_offense'
    test/rubocop/cop/my_cops/my_new_cop_test.rb:12:in `block in <class:MyNewCopTest>'
```

**After**

```
Error:
RuboCop::Cop::MyCops::MyNewCopTest#test_works:
RuntimeError: Cop not defined: RuboCop::Cop::MyCops::MyNewCop
    /Users/brandoncc/dev/rubocop-minitest/lib/rubocop/minitest/assert_offense.rb:81:in `cop'
    /Users/brandoncc/dev/rubocop-minitest/lib/rubocop/minitest/assert_offense.rb:111:in `assert_offense'
    test/rubocop/cop/my_cops/my_new_cop_test.rb:12:in `block in <class:MyNewCopTest>'
```

My first implementation [simply changed `AssertOffense#setup`](https://github.com/rubocop/rubocop-minitest/commit/8ab6c91e19a098eb4021f9afd0bbd35f2ebcaeb3) to fail if the cop wasn't found, but that broke four tests in the test suite. Given that the advice given is to require `support.rb`, which [includes `AssertOffense` on every instance of `Minitest::Test`](https://github.com/rubocop/rubocop-minitest/blob/4d08365be08d06b02bb2c11415a4e8060331a0bb/lib/rubocop/minitest/support.rb#L10), I realized that implementation was likely to break many gem-consumer repositories in this same way.

The second implementation refactors the initialization of a Cop to be lazy, so that the error is only raised when the cop is attempted to be accessed. This has the downside of `@cop` being `nil` until `#cop` is called, but these failures will only happen in tests where `@cop` is accessed directly. I believe these failures, if experienced by consumers of the gem, will be easier to resolve than the errors likely to have been introduced by my first implementation.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/minitest/assert_offense.rb
+++ b/lib/rubocop/minitest/assert_offense.rb
@@ -75,11 +75,13 @@ module RuboCop
     module AssertOffense
       private
 
-      def setup
-        cop_name = self.class.to_s.delete_suffix('Test')
-        return unless RuboCop::Cop::Minitest.const_defined?(cop_name)
+      def cop
+        @cop ||= begin
+          cop_name = self.class.to_s.delete_suffix('Test')
+          raise "Cop not defined: #{cop_name}" unless RuboCop::Cop::Minitest.const_defined?(cop_name)
 
-        @cop = RuboCop::Cop::Minitest.const_get(cop_name).new(configuration)
+          RuboCop::Cop::Minitest.const_get(cop_name).new(configuration)
+        end
       end
 
       def format_offense(source, **replacements)
@@ -95,7 +97,7 @@ module RuboCop
       def assert_no_offenses(source, file = nil)
         setup_assertion
 
-        offenses = inspect_source(source, @cop, file)
+        offenses = inspect_source(source, cop, file)
 
         expected_annotations = RuboCop::RSpec::ExpectOffense::AnnotatedSource.parse(source)
         actual_annotations = expected_annotations.with_offense_annotations(offenses)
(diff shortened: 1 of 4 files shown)


lib/rubocop/minitest/assert_offense.rb
```diff
@@ -105,8 +107,7 @@ def assert_no_offenses(source, file = nil)
 
       def assert_offense(source, file = nil, **replacements)
         setup_assertion
-
-        @cop.instance_variable_get(:@options)[:autocorrect] = true
+        enable_autocorrect
```
brandoncc, line 110: I was forced to make this change because changing the instance variable to a method call  caused method to have an `AbcSize` offense.

lib/rubocop/minitest/assert_offense.rb
```diff
@@ -75,11 +75,13 @@ module Minitest
     module AssertOffense
       private
 
-      def setup
-        cop_name = self.class.to_s.delete_suffix('Test')
-        return unless RuboCop::Cop::Minitest.const_defined?(cop_name)
+      def cop
+        @cop ||= begin
```
brandoncc, line 79: This is the meat of the change, lazily-loading the cop in order to raise a nice error when the cop is used if the class isn't available.

koic:
Thanks!

PR #322, opened 2024-09-26
Check for assertion method receiver - fixes #321

### LINKED ISSUE #321: MultipleAssertions false positive with 'look-alike' assertion methods
All methods which start with "assert_" or "refute_" are currently assumed to be (and counted as) Minitest assertions. E.g.

With `Minitest/MultipleAssertions` Max of 1 the following registers an offense.

```ruby
def test_asserts_once
  assert_equal(:assert_equal, Bar.assert_but_not_a_minitest_assertion)
end
```
--------

## Expected behavior

No offense should be registered, as only one Minitest assertion is called in the test.

## Actual behavior

Offense is registered.

See above code sample

## RuboCop version
```
$ [bundle exec] rubocop -V
1.66.1 (using Parser 3.3.5.0, rubocop-ast 1.32.3, running on ruby 3.3.5) [x86_64-linux]
  - rubocop-minitest 0.36.0
  - rubocop-performance 1.21.1
```

MatzFan:
Counting assertions based on a string prefix of "assert_" or "refute_" is replaced by a symbol comparison against the 38 Minitest assertion methods.

### DIFF
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'minitest/assertions'
+require 'minitest/mock' # includes Assertions#assert_mock
 require 'set'
 
 module RuboCop
@@ -10,7 +12,8 @@ module RuboCop
       include DefNode
       extend NodePattern::Macros
 
-      ASSERTION_PREFIXES = %w[assert refute].freeze
+      ASSERTION_REGEXP = /\A(assert|refute)/.freeze
+      MINITEST_ASSERTIONS = ::Minitest::Assertions.instance_methods.grep(ASSERTION_REGEXP)
 
       LIFECYCLE_HOOK_METHODS_IN_ORDER = %i[
         before_setup
@@ -105,10 +108,10 @@ module RuboCop
         return assertion_method?(node.expression) if node.assignment? && node.respond_to?(:expression)
         return false if !node.send_type? && !node.block_type? && !node.numblock_type?
 
-        ASSERTION_PREFIXES.any? do |prefix|
+        MINITEST_ASSERTIONS.any? do |assertion_method|
           method_name = node.method_name
 
-          method_name.start_with?(prefix) || node.method?(:flunk)
+          method_name == assertion_method || node.method?(:flunk)
         end
       end
       # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'minitest/assertions'
-require 'minitest/mock' # includes Assertions#assert_mock
 require 'set'
 
 module RuboCop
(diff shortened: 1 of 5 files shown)


Earlopain:
This is not the best solution. It will ignore all custom assertions the user may have written themselves, and test methods provided by gems like rails.

I think it would be better to instead only look at methods that have no receiver.

MatzFan:
@Earlopain 

> It will ignore all custom assertions the user may have written themselves, and test methods provided by gems like rails.

Apologies, I didn't see that tested for anywhere.

>I think it would be better to instead only look at methods that have no receiver.

Or where the receiver is not a kind_of `Minitest::Test`?

Earlopain:
> Apologies, I didn't see that tested for anywhere.

I guess the `assert_something` call you changed could be counted as that but yes, it isn't very explicit in the test suite yet. Let's just add a testcase for custom assertion methods with this PR.

> Or where the receiver is not a kind_of Minitest::Test?

Perhaps, but not decidable by rubocop. It can't tell if `MyCustomTestCase` or `AcitveSupport::TestCase` inherit from `Minitest::Test`. Just no receiver seems fine, unless you see a problem with that yourself.

-------

Please add a [changelog entry](https://github.com/rubocop/rubocop-minitest/blob/master/CONTRIBUTING.md#changelog-entry-format) for this, and write a meaningful commit message starting with `[Fix #321]`. Thanks!

MatzFan:
Will take a :eyes:

lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
```diff
@@ -108,10 +108,10 @@ def assertion_method?(node)
         ASSERTION_PREFIXES.any? do |prefix|
           method_name = node.method_name
 
-          method_name.start_with?(prefix) || node.method?(:flunk)
+          (method_name.start_with?(prefix) && !node.receiver) || node.method?(:flunk)
         end
       end
-      # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
```
koic, line 114: Can you refactor to the following?

```diff
diff --git a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
index 4bb7382..ceb1971 100644
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -99,19 +99,20 @@ module RuboCop
         end
       end

-      # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       def assertion_method?(node)
         return false unless node
+        return false if node.receiver
         return assertion_method?(node.expression) if node.assignment? && node.respond_to?(:expression)
         return false if !node.send_type? && !node.block_type? && !node.numblock_type?

         ASSERTION_PREFIXES.any? do |prefix|
           method_name = node.method_name

-          (method_name.start_with?(prefix) && !node.receiver) || node.method?(:flunk)
+          method_name.start_with?(prefix) || node.method?(:flunk)
         end
       end
-      # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
+      # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity

       def lifecycle_hook_method?(node)
         node.def_type? && LIFECYCLE_HOOK_METHODS.include?(node.method_name)
```

koic:
@MatzFan Can you rebase with the latest master branch and apply the above comment? Finally, please squash your commits into one.

MatzFan:
Sorry I dropped the ball on this one. I've just tried to finish it off but the addition of the guard `return false if node.receiver` in `lib/rubocop/cop/mixin/minitest_exploration_helpers.rb` now causes 2 failures. My original test still fails without this. Kinda stuck.

PR #335, opened 2025-05-12
Fix minitest extension for the cop generator

### LINKED ISSUE #334: `lib/rubocop/cop/generator.rb` can't be used in client code
## Expected behavior

Being able to load the `Rubocop::Cop::Generator` extensions defined in this library in a ruboocp plugin.

## Actual behavior

There is no way to require the `lib/rubocop/cop/generator.rb` since the file name matches the one in rubocop https://github.com/rubocop/rubocop/blob/6e9277dc18c0c2020301c8d18aa1c405bf462a47/lib/rubocop/cop/generator.rb.

This gem uses `require_relative` to solve that problem, but client code can't.

```
generator = RuboCop::Cop::Generator.new("SomeCop)

  generator.write_source
  generator.write_test
```

After loading this library.

rafaelfranca:
Instead of reopening the Generator class in a file with the same name of the Generator class in the main rubocop gem, we are now defining a module that get included in the Generator class, defined in a different file path, specific for this minitest plugin.

This allows rubocop plugins that want to use minitest to test their code to be able to generate their own cops using the minitest extension.

Fixes #334.

-----------------

* [ ] Added tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,7 +17,7 @@ end
 
 require 'rubocop/rake_task'
 require 'rake/testtask'
-require_relative 'lib/rubocop/cop/generator'
+require 'rubocop/cop/minitest_generator'
 
 # JRuby and Windows don't support fork, so we can't use minitest-queue.
 if RUBY_ENGINE == 'jruby' || RuboCop::Platform.windows?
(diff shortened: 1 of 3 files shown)


koic:
Thanks!

PR #342, opened 2025-10-30
[Fix #320] Add cop for test class name

### LINKED ISSUE #320: [Enhancement] Add cop for test class name

Presently, Rubocop expects a test class name to end with "Test" and this is hard coded at [this line](https://github.com/rubocop/rubocop-minitest/blob/4d08365be08d06b02bb2c11415a4e8060331a0bb/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb#L29). As per [this comment](https://github.com/rubocop/rubocop-minitest/issues/312#issuecomment-2229076243) using test class names like `TestFoo` (which happens to be the [convention used](https://github.com/minitest/minitest?tab=readme-ov-file#unit-tests-) by Minitest itself) results in Rubocop ignoring the methods in that test class entirely.

I'd like to see:

1. A new cop to detect where a test file includes a class which includes test methods but is not named according to the expected convention (i.e. class names not ending with "Test"). Note it would be important to ignore classes in test files with no test methods, so as to avoid false positives. It is easy (I've done it myself) to name a test class `Foo` (i.e. mistakenly omit the "Test" suffix) or to incorrectly name a test class with a typo - e.g. `FooTst`. In such cases all test methods included in the class are skipped by Rubocop, though the code will still run.
2. Perhaps also the ability to configure the convention hard coded above such that a project could use the "TestFoo" convention instead. I'm not a Rails dev and was not aware of the Rails convention and I guess the demand for anything other than the "FooTest" convention may be vanishingly small, so this requirement may be unnecessary.

## Describe alternatives you've considered

Leave as is and expect users to correctly name their test classes!

Magikdidi24:
# Fix test class detection to avoid false positives

## Summary

This PR fixes the test class detection logic in `MinitestExplorationHelpers` to eliminate false positives and false negatives. The detection now relies **solely on the parent class** rather than class naming conventions.

## Problem

The previous implementation had two major issues:

### 1. False Positives
Classes with "Test" in their name but not inheriting from test frameworks were incorrectly identified as test classes:

```ruby
class FooTest < ApplicationRecord  # Incorrectly identified as test class
  def perform; end
end

class ModelFooTest < ActiveJob::Base  # Incorrectly identified as test class
  def perform; end
end
```

This caused RuboCop to run Minitest cops on non-test code.

### 2. False Negatives
Classes following Minitest's `TestFoo` naming convention were not recognized:

```ruby
class TestFoo < Minitest::Test  # Not recognized as test class
  def test_something
    assert true
  end
end
```

This caused RuboCop to skip checking legitimate test classes.

### Usage Example

**Project with custom test base class:**

```ruby
# test/support/parent_test.rb
class ParentTest < Minitest::Test
  def setup
    # Common setup
  end
end

# test/models/user_test.rb
class UserTest < ParentTest
  def test_something
    assert true
  end
end
```

**Configuration needed:**
```yaml
# .rubocop.yml
Minitest:
  AdditionalTestBaseClasses:
    - ParentTest
```

Without this configuration, `UserTest < ParentTest` will not be recognized as a test class.

**Migration guide for users:**

If you have custom test base classes like:

```ruby
class ParentTest < Minitest::Test; end
class MyCustomBase < Minitest::Test; end
```

And your test classes inherit from them:

```ruby
class UserTest < ParentTest; end
class FooTest < MyCustomBase; end
```

Add this to your `.rubocop.yml`:

```yaml
Minitest:
  AdditionalTestBaseClasses:
    - ParentTest
    - MyCustomBase
```

## Related Issues

Fixes issue mentioned in comments about:
- Classes named `FooTest < ApplicationRecord` being incorrectly identified as test classes
- `TestFoo` naming convention not being recognized
- Need for configurable test base classes

## Documentation

Users should be informed about the `AdditionalTestBaseClasses` configuration option for custom test base classes.
Related to #320 

-----------------

* [x] Added tests.
*  Added an entry (file) to the [changelog folder](https://github.com/rubocop/rubocop-minitest/blob/master/changelog/) named `{change_type}_{change_description}.md` if the new code introduces user-observable changes. See [changelog entry format](https://github.com/rubocop/rubocop/blob/master/CONTRIBUTING.md#changelog-entry-format) for details.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4,6 +4,9 @@ Minitest:
   Include:
     - '**/test/**/*'
     - '**/*_test.rb'
+  # Additional test base class names to recognize as test classes.
+  # Example: ['MyCustomBase', 'ApplicationTestCase']
+  AdditionalTestBaseClasses: []
 
 Minitest/AssertEmpty:
   Description: 'This cop enforces the test to use `assert_empty` instead of using `assert(object.empty?)`.'
--- a/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
+++ b/lib/rubocop/cop/mixin/minitest_exploration_helpers.rb
@@ -6,7 +6,7 @@ module RuboCop
   module Cop
     # Helper methods for different explorations against test files and test cases.
     # @api private
-    module MinitestExplorationHelpers
+    module MinitestExplorationHelpers # rubocop:disable Metrics/ModuleLength
       include DefNode
       extend NodePattern::Macros
 
@@ -40,8 +40,29 @@ module RuboCop
 
       private
 
+      def additional_test_base_classes
+        return [] unless respond_to?(:cop_config)
+
+        cop_config.fetch('AdditionalTestBaseClasses', [])
+      rescue StandardError
+        []
+      end
+
+      def test_base_classes
+        default_base_classes = %w[
+          Minitest::Test
+          ActiveSupport::TestCase
+          ActionController::TestCase
+          ActionDispatch::IntegrationTest
+        ]
+
+        default_base_classes + additional_test_base_classes
+      end
+
       def test_class?(class_node)
-        class_node.parent_class && class_node.identifier.source.end_with?('Test')
+        return false unless class_node.parent_class
+
+        test_base_classes.include?(class_node.parent_class.source)
       end
 
       def test_case?(node)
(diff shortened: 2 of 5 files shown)


Magikdidi24:
Hey, @koic , @flavorjones, Any issue with this PR?
The detection based on parent classes is more accurate than using naming rules and avoids both false positives and false negatives. The AdditionalTestBaseClasses option covers custom setups. Let me know if changes are needed.

PR #355, opened 2026-07-12
Add `RuboCop::TestCase` as a stable API for testing custom cops

koic:
This gem has provided `assert_offense` and friends for custom cop development since v0.15, but only as a mixin into `Minitest::Test`. Some helpers such as `inspect_source` required passing the cop instance explicitly, and there was no natural way to specify the cop configuration or the target Ruby version.

Introduce `RuboCop::TestCase`, a base test case class named in the same manner as `ActiveSupport::TestCase`, as the stable public API:

- The cop under test is derived from the test class name (e.g. `CustomCops::MyCopTest` resolves `CustomCops::MyCop`, then `RuboCop::Cop::Minitest::MyCop`) and can be specified explicitly by defining a `cop_class` method.
- `inspect_source(source, file = nil)` and `investigate(processed_source)` no longer take a cop argument; they use the cop under test. (Breaking)
- `cop_config`, `other_cops`, and `target_ruby_version` can be overridden in the test class, mirroring the RSpec `let` based configuration of rubocop/rubocop specs in a way that is natural for Minitest. Assigning them in a test method changes them for that test only, so tests targeting different Ruby versions can live in the same test class. `target_ruby_version` flows into both the parser and `AllCops: TargetRubyVersion`. All of this state lives in the per-test instance and the one-time plugin integration is guarded by a mutex, so the assertions are safe under multithreaded parallel execution such as `parallelize_me!`.
- The configuration merges the default configuration of the cop under test, integrating lint_roller plugin defaults, in parity with the RSpec `config` shared context of rubocop/rubocop.

The existing mixin into `Minitest::Test` via `require 'rubocop/minitest/support'` is kept for backward compatibility.

Since the cop and its configuration are now built lazily per test instance, this also removes the hand-built `@cop` hacks from this gem's own tests and unskips the tests that were marked flaky because of the shared `@cop` state. The todo generation test is rewritten against the `RuboCop::ExcludeLimit` tmp file tracking API, which replaced `config_to_allow_offenses[:exclude_limit]` in RuboCop core.

-----------------

* [x] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,96 @@ Minitest/AssertNil:
     - test/my_file_to_ignore_test.rb
 ```
 
+## Testing your own cops
+
+This gem provides `RuboCop::TestCase`, a base test case class for testing custom cops with Minitest.
+Require `rubocop/minitest/support` in your test helper, then inherit from `RuboCop::TestCase` to get
+the `assert_offense`, `assert_correction`, `assert_no_corrections`, and `assert_no_offenses` assertions:
+
+```ruby
+require 'rubocop/minitest/support'
+
+module CustomCops
+  class MyCopTest < RuboCop::TestCase
+    def test_registers_offense
+      assert_offense(<<~RUBY)
+        bad_method
+        ^^^^^^^^^^ Use `good_method` instead of `bad_method`.
+      RUBY
+
+      assert_correction(<<~RUBY)
+        good_method
+      RUBY
+    end
+
+    def test_does_not_register_offense
+      assert_no_offenses(<<~RUBY)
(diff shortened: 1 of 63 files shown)


rafaelfranca (commented):
Amazing. I have a lot of projects that I had to implement this. Thank you for working on it

andyw8:
@koic btw the [release notes for v0.40.0](https://github.com/rubocop/rubocop-minitest/releases/tag/v0.40.0) are linking to the wrong PR (#278 instead of this).

exterm:
😍

PR #363, opened 2026-08-14
Optimize `MultipleAssertions` cop

moberegger:
I'm trying to trim down my Rubocop run times in a rather large project. The `Minitest/MultipleAssertions` came up as one of most expensive ones in our profiles. It took 6.74 seconds across  4,261 calls (about 1.6 ms per call).

There are three optimizations in this PR:
- Makes `MATCHER_METHODS` a `Set`. This makes the `MATCHER_METHODS.include?` call in `assertion_method?` an O(1) operation. There are other cops using this helper method, so this may improve performance of those, too.
- Moves the `when *RuboCop::AST::Node::ASSIGNMENTS` to be an inline `RuboCop::AST::Node::ASSIGNMENTS.include?(node.type)` in the fallback branch of `assertions_count_based_on_type`'s `case` expression. `RuboCop::AST::Node::ASSIGNMENTS` is a `Set`, and splatting it turns it into an `Array`, which causes any node that falls through to do an O(n) lookup. The proposed optimization keeps it O(n)
- In the same branch, perform a manual tally instead of using `node.each_child_node.sum`. The latter results in an `Enumerator` being allocated for each node, which slows things down. Doing a manual tally saves on the allocations.

I've included a benchmark below with each proposed change run in isolation, and a final one with all three changes applied. 15 passes were done, with each pass being 600 cop executions. Every version registered 7,200 offenses.

| version                      | mean   | min    | max    | vs original |
|------------------------------|--------|--------|--------|-------------|
| original                     | 0.929s | 0.911s | 0.945s | 1.00x       |
| MATCHER_METHODS as a Set     | 0.842s | 0.826s | 0.880s | 1.10x       |
| ASSIGNMENTS out of `when`    | 0.831s | 0.807s | 0.873s | 1.12x       |
| each_child_node with a block | 0.811s | 0.790s | 0.845s | 1.15x       |
| all three                    | 0.621s | 0.595s | 0.660s | 1.50x       |

-----------------

* [ ] Added tests.

### DIFF
--- a/changelog/change_improve_multiple_assertions_performance.md
+++ b/changelog/change_improve_multiple_assertions_performance.md
@@ -1,0 +1,1 @@
+* [#363](https://github.com/rubocop/rubocop-minitest/pull/363): Improve `Minitest/MultipleAssertions` performance by removing per-node allocations and a linear matcher-method lookup from the assertion counter. ([@moberegger][])
--- a/lib/rubocop/cop/minitest/multiple_assertions.rb
+++ b/lib/rubocop/cop/minitest/multiple_assertions.rb
@@ -67,10 +67,15 @@ module RuboCop
             assertions_count(node.body) + assertions_count_in_branches(node.branches)
           when :block, :numblock, :itblock
             assertions_count(node.body)
-          when *RuboCop::AST::Node::ASSIGNMENTS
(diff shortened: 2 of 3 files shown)


koic:
Thanks!

PR #365, opened 2026-09-08
[Fix #310] Don't autocorrect `AssertMatch`/`RefuteMatch` when the argument order is ambiguous

### LINKED ISSUE #310: `Minitest/AssertMatch` autofix can be incorrect
When applying `Minitest/AssertMatch`'s auto-fix to the following the working code:

```ruby
require "minitest/autorun"

class AssertMatchTest < Minitest::Test
  def test_assert_match_autofix
    hello = "Hello World!"
    greeting = %r{Hello}
    assert hello.match?(%r{World})
    assert hello.match? greeting
  end
end
```

## Expected behavior

It would auto-correct `assert hello.match? greeting` to `assert_match greeting, hello`.
Indeed, it works correctly for `assert hello.match?(%r{World})`.

## Actual behavior

It auto-corrects `assert hello.match? greeting` to `assert_match hello, greeting` (note the order of parameters), which causes the test to error:

```
TypeError: no implicit conversion of Regexp into String
```

since it didn't swap the order of the parameters.

1. Save the above code into a file (`foo_test.rb`)
2. Run the test to verify that it passes (`ruby foo_test.rb`)
3. Autoformat the code (`rubocop -a foo_test.rb`)
4. Run the test again (`ruby foo_test.rb`)

See the test error.

## RuboCop version

```
➜ bundle exec rubocop -V
1.64.1 (using Parser 3.3.0.5, rubocop-ast 1.31.2, running on ruby 3.1.4) [arm64-darwin23]
  - rubocop-minitest 0.35.0
  - rubocop-performance 1.21.0
```

pcbeingused333:
Fixes #310.

### Problem

`a.match?(b)` accepts either operand as the pattern — `String#match?` and `Regexp#match?` both exist — so rewriting it into `assert_match(matcher, string)` order is a guess whenever neither operand is a regexp literal. The cop assumed the receiver was the pattern:

```ruby
def order_expected_and_actual(expected, actual)
  if actual.regexp_type?
    [actual, expected]
  else
    [expected, actual]   # <- receiver assumed to be the matcher
  end.map(&:source).join(', ')
end
```

So the working test

```ruby
hello = "Hello World!"
greeting = /Hello/
assert hello.match?(greeting)
```

autocorrected to `assert_match hello, greeting` — arguments in string/regexp order — which raises `TypeError: no implicit conversion of Regexp into String` at run time.

### Fix

The offense is still reported (the `assert_match` / `refute_match` suggestion is valid advice), but the **autocorrection is applied only when one operand is a regexp literal**, so the `(matcher, string)` order is actually known. When neither is a literal, the offense is reported without a correction.

`#323` proposed an opt-in config to silence the offense entirely; this instead keeps the offense and only drops the unsafe rewrite.

### Tests

- New `test_does_not_correct_when_neither_operand_is_a_regexp_literal` in both cop tests — reproduces #310 and asserts no correction.
- The existing `matcher.match?(object)` / `matcher =~ object` / `assert_operator(matcher, :=~, object)` cases (bare receiver and argument) now assert `assert_no_corrections` instead of a correction.
- The regexp-literal cases (`/regexp/.match?(x)`, `x.match?(/regexp/)`) are unchanged.

`bundle exec rake` is green — 2520 runs, 0 failures; internal RuboCop clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pcbeingused333:
The workflows are on `action_required` (fork gate) — could a maintainer press "Approve and run workflows"? Local verification:

- `bundle exec rake` — green: 2520 runs, 0 failures; internal RuboCop clean (142 files, no offenses); changelog/doc checks pass.
- The three new "does not correct" assertions (`assert`/`refute` variants of the #310 repro, plus the `assert_operator`/`refute_operator` ambiguous forms) fail on `master` — they show the old wrong-order autocorrect.

PR #366, opened 2026-09-09
[Fix] Fix an error for `Minitest/GlobalExpectations` on a nested matcher receiver

pcbeingused333:
## Summary

`Minitest/GlobalExpectations` raises `Parser::ClobberingError` during autocorrection when the receiver of a matcher call is itself another matcher call:

```ruby
a.must_equal(b).must_equal(c)
```

## Problem

Both `must_equal` calls are matcher calls, so both get an offense. The inner one wraps `a` as `_(a)`; the outer one wraps its own receiver (`a.must_equal(b)`) as `_(a.must_equal(b))`. The outer replacement range fully contains the inner one — the corrector clobbers when it tries to apply both in the same pass.

## Fix

Skip the correction when the node about to be wrapped is itself a matcher call — it is already offended and will be wrapped by its own correction. The offense is still reported (unfixed), matching how the cop already leaves other shapes uncorrected rather than guessing.

## Tests

- New `test_does_not_crash_when_receiver_is_itself_a_matcher_call`, run under all 4 `EnforcedStyle` subclasses (`any`/`_`/`expect`/`value`) — 4 runs, 8 assertions.
- Verified in red: reverting the fix reproduces the crash in all 4 (`Parser::ClobberingError`).
- `bundle exec rake test`: 2522 runs, 8534 assertions, 0 failures.
- `bundle exec rake internal_investigation`: 142 files, no offenses.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PjzCQWswbH69CXJ2nCa9Zk


pcbeingused333:
The workflows are on `action_required` (fork gate) — could a maintainer press "Approve and run workflows"?

Local verification while waiting: `bundle exec rake test` (2522 runs, 8534 assertions, 0 failures), `bundle exec rake internal_investigation` (142 files, no offenses), `bundle exec ruby -Ilib -Itest test/project_test.rb` (changelog format, 18 runs, 0 failures).

koic:
Thanks for tracking this down. The bug is real, but I went with a different fix.

Skipping the outer correction regresses autocorrection: RuboCop's correction loop already recovers from the clobbering, so master reaches `_(_(a).must_equal(b)).must_equal(c)`, while this stops at `_(a).must_equal(b).must_equal(c)` and leaves an offense `rubocop -a` can never clean up. It also misses `_(a.must_equal(b)).must_equal(c)` under `EnforcedStyle: expect`, since `matcher_call?` does not match `_(...)`.

Instead of replacing the whole receiver, #367 replaces only the DSL method selector and uses `corrector.wrap` for the new calls, so the ranges nest instead of overlapping. Opened with you as a co-author, thanks again for the feedback!

